Issue #90, opened 2026-05-05
feat(journal): add tier tagging to auto-log hooks (prerequisite for Tier Summary)


danielbentes:
## Context

Decision-journal entries auto-logged by hooks today contain action type only, not the three-tier classification:

```
<!-- auto-log: 2026-05-05 23:10 Edit /path/to/file.md -->
<!-- auto-log: 2026-05-05 23:10 commit "fix(skill): ..." -->
```

There are no Tier 2 or Tier 3 events captured at all (push, PR create, issue assign, merge, release). The `log-file-changes.sh` and `log-commits.sh` PostToolUse hooks cover Tier 1 actions (Edit, Write, commit) but emit no tier metadata. There are no hooks for Tier 2/3 actions.

The Tier Summary feature in `/flow:status` (the desired output is shown in the workshop slide mockup at `docs/flow-team-session/slides.md` lines 802-810) needs three-tier event counts. Without tier-tagged journal entries AND hooks that capture Tier 2/3 events, that feature cannot accurately match its slide mockup.

This issue blocks the Findings Ledger and Tier Summary work.

## Current State

- `plugins/flow/hooks/scripts/log-file-changes.sh` (PostToolUse Edit|Write): emits `<!-- auto-log: TS Edit /path -->` — no tier tag
- `plugins/flow/hooks/scripts/log-commits.sh` (PostToolUse Bash, matches `git commit`): emits `<!-- auto-log: TS commit \"msg\" -->` — no tier tag
- No hook captures: `git push`, `gh pr create`, `gh pr merge`, `gh release create`, `gh issue edit --add-assignee`
- `plugins/flow/skills/autonomous-workflow/SKILL.md` documents the three tiers conceptually but not a journal tag convention

## Objective

Auto-logged journal entries carry an explicit tier tag (T1, T2, or T3), and hook coverage extends to include Tier 2 (push, PR create, issue assign) and Tier 3 (merge, release) actions, so a downstream consumer (Tier Summary in `/flow:status`) can aggregate counts that match the slide mockup.

## Acceptance Criteria

- [ ] `log-file-changes.sh` emits journal entries that include a `T1` tag (e.g. `<!-- auto-log: TS T1 Edit /path -->`).
- [ ] `log-commits.sh` emits journal entries that include a `T1` tag.
[...]

Issue #85, opened 2026-05-05
feat(status): add Tier Summary section to /flow:status


danielbentes:
## Context

The workshop deck shows a `/flow:status` mockup that includes a `### Tier Summary` section reporting Tier 1/2/3 event counts during the current session/period. This signals workflow drift (e.g., too many Tier 3 events suggests unsafe patterns) and is described as if it ships today (slides.md lines 802-810). The actual `commands/status.md` does not generate it.

This issue likely depends on a journal-schema decision: do existing `auto-log` markers carry tier metadata? If not, this becomes a 2-step issue chain — first add tier-tagging to journal entries, then aggregate counts in status.

## Current State

- `plugins/flow/commands/status.md` does not produce a Tier Summary section.
- The slide mockup (lines 802-810) shows the desired format:
  ```
  ### Tier Summary
  Tier 1 events: 47 (commits, branches, tests)
  Tier 2 events: 6 (pushes, PR creates)
  Tier 3 events: 0 (no merges or releases this session)
  ```
- The journal directory `.decisions/` accumulates entries from `log-file-changes.sh` and `log-commits.sh` hooks plus manual journal writes. Whether entries carry a tier tag needs verification.

## Objective

A user running `/flow:status` sees a Tier Summary section that aggregates Tier 1/2/3 event counts from the journal, surfacing workflow drift signals.

## Acceptance Criteria

- [ ] `/flow:status` output includes a `### Tier Summary` section in the position shown by the slide mockup (after Decision Journal, before Suggested Next Action — to confirm).
- [ ] The section shows three counts: Tier 1, Tier 2, Tier 3, each with example actions in parentheses.
- [ ] Counts are sourced from journal entries (under `.decisions/` or as configured by `journal.dir`).
- [ ] If existing journal entries do not carry tier metadata, a sub-issue is filed first to add tier-tagging to `log-file-changes.sh`, `log-commits.sh`, and the manual-write protocol in `autonomous-workflow/SKILL.md`. This issue is then re-scoped to consume that schema.
- [ ] Empty state ("no journal entries this session") renders cleanly.
- [ ] The output matches the format shown in `docs/flow-team-session/slides.md` lines 802-810.

## Verification

```text
# 1. Inspect existing journal entries for tier metadata
grep -h "tier\|Tier" .decisions/*.md
# Expected (post-implementation): each entry has a tier tag

# 2. Empty-journal state — in a test fixture with cleared .decisions
/flow:status
[...]

Issue #84, opened 2026-05-05
feat(status): add Findings Ledger section to /flow:status


danielbentes:
## Context

The workshop deck shows a `/flow:status` mockup that includes a `### Findings Ledger` section reporting P1/P2/P3 finding counts across open PRs. This is a useful daily-driver for spotting blocked merges and pending escalations. The section is described as if it ships today (slides.md lines 802-810), but the actual `commands/status.md` does not generate it. Workshop attendees who run `/flow:status` will see a different output than the slide.

The infrastructure exists: `commands/merge.md` lines 48-76 already parses `<!-- FLOW_RESOLUTION_CYCLE:N RESOLVED:[...] ESCALATED:[...] DISPUTED:[] -->` markers to verify merge prerequisites. Reusing that parser across all open PRs surfaces the ledger.

## Current State

- `plugins/flow/commands/status.md` produces sections: Current Branch, My Issues (Open), My PRs, Awaiting My Review, Decision Journal, Suggested Next Action.
- It does NOT produce a Findings Ledger.
- The slide mockup at lines 802-810 shows the desired format:
  ```
  ### Findings Ledger
  P1: 0    P2: 1 (in fix-forward)    P3: 1 (ESCALATED — awaiting reviewer accept)
  ```
- FLOW_RESOLUTION_CYCLE marker format is defined by `templates/resolution-comment.md` and parsed by `merge.md` lines 48-76.

## Objective

A user running `/flow:status` sees a Findings Ledger section that summarizes open-PR findings by priority, sourced from FLOW_RESOLUTION_CYCLE markers, matching the format shown in the workshop slide.

## Acceptance Criteria

- [ ] `/flow:status` output includes a `### Findings Ledger` section after the existing sections (or in the position shown by the slide mockup).
- [ ] The section shows P1, P2, P3 counts across all open PRs assigned to or authored by the current user.
- [ ] Counts are sourced from FLOW_RESOLUTION_CYCLE markers in PR comments (reuses the parsing logic from `merge.md` lines 48-76 — extract into shared bash function or skill reference).
- [ ] When a P2 finding is in fix-forward state, it is annotated as such (e.g., "P2: 1 (in fix-forward)").
- [ ] When a P3 finding is ESCALATED, it is annotated as such (e.g., "P3: 1 (ESCALATED — awaiting reviewer accept)").
- [ ] Empty state ("no open PRs with findings") renders cleanly without producing an empty section or error.
[...]

PR #91, opened 2026-05-05
feat(journal): add tier tagging to auto-log hooks

danielbentes:
## Summary

Auto-logged journal entries now carry an explicit tier tag (T1/T2/T3) so downstream consumers (`/flow:status` Tier Summary) can aggregate events by tier. Hook coverage extended from T1-only (file edits, commits) to also include T2 (push, PR create, issue assign) and T3 (merge, release).

This is the prerequisite for the combined Findings Ledger + Tier Summary status enhancements.

## Comprehension Report

A spike on the existing journal entries showed every auto-log marker had only `<TIMESTAMP> <ACTION> <TARGET>` — no tier classification, and zero coverage for Tier 2/3 events (push, PR create, merge, release, issue assign). Without tier-tagged entries AND hooks for the relevant Tier 2/3 actions, the workshop slide mockup of `/flow:status` (which shows `Tier 2 events: 6 (pushes, PR creates)` etc.) couldn't be implemented accurately. This PR is the foundation: T1/T2/T3 tags emitted by hooks now match the slide's expected output shape.

The new format is additive — older un-tagged entries remain parseable. SKILL.md documents the canonical parser regex `<!-- auto-log: ([0-9-]+ [0-9:]+) (?:(T[123]) )?(.+) -->` so consumers can match both formats with one pattern; un-tagged entries default to T1 (the only tier that was tracked before tier tagging shipped).

## Per-AC Verdict (all PASS)

Independent verdict-judge run with the new Read-only budget — 8/8 PASS:

| # | Acceptance Criterion | Verdict | Evidence |
|---|---|---|---|
| 1 | log-file-changes.sh emits T1 tag | PASS | `<!-- auto-log: TS T1 Edit /path -->` produced from sample stdin |
| 2 | log-commits.sh emits T1 tag | PASS | `<!-- auto-log: TS T1 commit \"msg\" -->` produced |
| 3 | T2 hook coverage (push, pr-create, issue-assign) | PASS | All 3 produced T2-tagged entries from sample stdin |
| 4 | T3 hook coverage (merge, release) | PASS | Both produced T3-tagged entries; mixed-tier `gh pr merge && git push` correctly tagged T3 (highest-tier wins) |
| 5 | SKILL.md documents tier tag convention | PASS | `Tier tag convention` section with format spec, hook coverage table, and parser-regex example |
| 6 | Backwards-compat: un-tagged entries still parseable | PASS | Single regex matches both formats across all 4 existing journal files |
| 7 | Idempotency: chore-commit + journal-only guards intact | PASS | Both guards byte-identical to main; new hook is loop-safe by design |
| 8 | hooks.json registers new hook | PASS | `log-tier-events.sh` registered under PostToolUse Bash matcher; valid JSON |

## Files changed

- `plugins/flow/hooks/scripts/log-file-changes.sh` — emit `T1` between timestamp and action
- `plugins/flow/hooks/scripts/log-commits.sh` — emit `T1`; tightened command-detection regex
- `plugins/flow/hooks/scripts/log-tier-events.sh` — NEW; PostToolUse Bash hook detecting T2/T3 commands
- `plugins/flow/hooks/hooks.json` — register the new hook
- `plugins/flow/skills/autonomous-workflow/SKILL.md` — Tier tag convention subsection
- `.decisions/issue-90.md` — decision journal

## Self-review findings

Cycle 1 surfaced 6 P2 findings. **All fixed-forward** in commit `7585eb9`:

- **P2#1** — Quoted strings could false-match. `git commit -m 'gh pr merge soon'` was tagged T3. Fix: strip quoted strings via sed before pattern matching.
- **P2#2** — Header comment said \"first match wins\" but implementation is highest-tier-wins. Comment corrected.
- **P2#3** — `\\b` after keywords matched `push-foo`. Replaced with explicit end-class `([[:space:]]|$|[^a-zA-Z0-9_-])`.
- **P2#4** — `$()` and backtick subshells escaped detection. Extended leading-anchor class to include `(` and backtick.
- **P2#5** — log-commits.sh had the same loose regex. Tightened to match.
- **P2#6** — Backwards-compat instruction in SKILL.md was buried. Added explicit parser-regex example.

P1: 0. P3 actionable: 0.

## Holdout validation

PASS — 8 holdout scenarios verified. No conflicts between self-review claims and file state. Notably:
- Quote-stripping verified for both single and double quotes
- End-class refuses hyphenated false positives (`push-foo`, `pushabc`)
- Leading-anchor includes `(` and backtick for subshell detection
- log-commits.sh chore-commit + journal-only guards byte-identical to main
- Parser regex valid PCRE; matches both old and new formats

## Test plan

- [x] All 8 ACs verified with sample stdin per command type
- [x] Edge cases: false-positive guards on `gh pr view`, `gh release list`, `gh issue view` (no log emitted)
- [x] Edge case: highest-tier wins on mixed `gh pr merge && git push` (T3 logged)
- [x] Edge case: quoted commit messages with trigger words don't false-match (post-fix-forward)
- [x] Edge case: `$(gh pr merge ...)` subshell wrappers detected (post-fix-forward)
- [x] Self-review: 6 P2 findings surfaced and fix-forwarded
- [x] Holdout validation: 8 scenarios PASS
- [x] Verdict-judge: 8/8 ACs PASS
- [x] Static checks: bash syntax clean, hooks.json valid JSON, new hook executable
- [ ] Reviewer reads `log-tier-events.sh` quote-stripping + anchor logic and confirms regex correctness
- [ ] Reviewer confirms backwards-compat parser regex in SKILL.md works for the existing un-tagged entries

## Followups

Once this lands, the combined `/flow:status` Findings Ledger + Tier Summary PR becomes implementable with accurate tier counts.

### DIFF
--- a/.decisions/issue-90.md
+++ b/.decisions/issue-90.md
@@ -85,3 +85,21 @@ Tasks are independent (no shared file beyond hooks.json which is touched only by
 <!-- auto-log: 2026-05-06 00:41 commit "fix(verdict-judge): tighten tool budget to Read-only (#89)" -->
 
 <!-- auto-log: 2026-05-06 00:41 T3 merge -->
+
+<!-- auto-log: 2026-05-06 00:42 commit "feat(journal): add tier tagging to auto-log hooks" -->
+
+<!-- auto-log: 2026-05-06 00:43 commit "feat(journal): add tier tagging to auto-log hooks" -->
+
+<!-- auto-log: 2026-05-06 00:45 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/hooks/scripts/log-tier-events.sh -->
+
+<!-- auto-log: 2026-05-06 00:45 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/hooks/scripts/log-commits.sh -->
+
+<!-- auto-log: 2026-05-06 00:45 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/skills/autonomous-workflow/SKILL.md -->
+
+<!-- auto-log: 2026-05-06 00:46 T1 commit "feat(journal): add tier tagging to auto-log hooks" -->
+
+<!-- auto-log: 2026-05-06 00:46 T3 merge -->
+
+<!-- auto-log: 2026-05-06 00:46 T2 push -->
+
+<!-- auto-log: 2026-05-06 00:46 commit "feat(journal): add tier tagging to auto-log hooks" -->
--- a/.decisions/issue-90.md
+++ b/.decisions/issue-90.md
@@ -1,0 +1,87 @@
+# Decision Journal — Issue #90
+
+**Title**: feat(journal): add tier tagging to auto-log hooks (prerequisite for Tier Summary)
+**Branch**: feature/issue-90-journal-tier-tagging
+**Started**: 2026-05-06
+
+## Specification
+
+### Non-goals
+- Don't change journal-file naming or directory configuration logic.
+- Don't modify the three blocking hooks (block-force-push, block-destructive, block-secrets).
+- Don't break the existing chore-commit guard or journal-only guard in log-commits.sh.
+- Don't add tier tagging for tools outside the explicit T2/T3 list (push, PR create, issue assign, merge, release).
+- Don't change the journal entry naming convention (`issue-N.md` or `session-YYYY-MM-DD.md`).
+
+### Failure modes
+- **jq unavailable** → graceful exit 0 (existing pattern in log-* hooks).
+- **Journal file missing** → skip silently (existing pattern).
+- **Detached HEAD or no branch** → fall back to `session-{date}.md` (existing pattern).
+- **Mixed-tier command** (e.g. `git push && gh pr create`) → tag the first detected match. Document this as a known limitation.
+- **False positives on non-targeted gh commands** (e.g. `gh pr view`, `gh pr list`, `gh release list`) → regex must specifically match `create`/`merge` actions, not arbitrary substrings.
+- **New hook fires on its own journal write** → hook is PostToolUse Bash; it only matches command-execution events, not file writes. Self-recursion not possible.
+
+### Interface contracts
+- **Journal entry format (new)**: `<!-- auto-log: YYYY-MM-DD HH:MM T{1,2,3} {action} {target} -->`
+- **Backwards-compat format (old)**: `<!-- auto-log: YYYY-MM-DD HH:MM {action} {target} -->` — downstream consumers must accept both; un-tagged entries default to T1.
+- **hooks.json schema**: unchanged (existing JSON shape, new entry under PostToolUse Bash matcher).
+- **Hook input contract**: stdin JSON with `tool_name` and `tool_input.command`/`tool_input.file_path` — same as existing hooks.
+
+## Spec Validation Gate
+
+| # | Acceptance Criterion | Verification Command | Gate Status |
+|---|---|---|---|
+| 1 | log-file-changes.sh emits T1 tag | `echo '{"tool_name":"Edit","tool_input":{"file_path":"/tmp/x"}}' \| bash plugins/flow/hooks/scripts/log-file-changes.sh && grep "T1 Edit" .decisions/*.md` | PASS |
+| 2 | log-commits.sh emits T1 tag | `echo '{"tool_input":{"command":"git commit -m test"}}' \| bash plugins/flow/hooks/scripts/log-commits.sh && grep "T1 commit" .decisions/*.md` | PASS |
+| 3 | New hook captures T2 events | `echo '{"tool_input":{"command":"git push origin x"}}' \| bash plugins/flow/hooks/scripts/log-tier-events.sh && grep "T2 push" .decisions/*.md` | PASS |
+| 4 | New hook captures T3 events | `echo '{"tool_input":{"command":"gh pr merge 88 --squash"}}' \| bash plugins/flow/hooks/scripts/log-tier-events.sh && grep "T3 merge" .decisions/*.md` | PASS |
+| 5 | autonomous-workflow/SKILL.md documents tier tag convention | `grep -A2 "tier tag" plugins/flow/skills/autonomous-workflow/SKILL.md` returns the convention section | PASS |
+| 6 | Backwards-compat: old format still parses | `grep -E "auto-log: [0-9-]+ [0-9:]+ (T[123] )?(Edit\|Write\|commit\|push\|create\|merge\|release\|assign)" .decisions/*.md \| wc -l` matches every auto-log line | PASS |
+| 7 | Idempotency: chore-guard intact, no log loops | `git diff plugins/flow/hooks/scripts/log-commits.sh` shows the chore-commit and journal-only guards unchanged | PASS |
+| 8 | hooks.json registers new hook | `jq '.hooks.PostToolUse[] \| select(.matcher == "Bash") \| .hooks[] \| .command' plugins/flow/hooks/hooks.json` includes log-tier-events.sh | PASS |
+
+All ACs PASS.
+
+## Stranger Test
+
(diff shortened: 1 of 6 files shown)


danielbentes (commented):
## Self-Review Cycle 2

Cycle-2 dispatched 3 parallel reviewers (code-reviewer, security-reviewer, error-handler-inspector) against the post-fix-forward state at commit 4881bcc.

### Findings: P1: 3, P2: 9, P3: 8

This is a **convergence-failure signal** per the autonomous-workflow stop conditions: cycle-1 surfaced 6 P2; cycle-2 surfaces 3 P1 + 9 P2 + 8 P3. The text-pattern approach to bash command parsing is fundamentally too fragile for a hook that runs on every Bash call.

### P1 — Critical (Blocks Merge)

| # | Source | Location | Issue |
|---|---|---|---|
| 1 | code-reviewer | `log-tier-events.sh` quote-strip sed | Cycle-1 P2 fix incomplete: an apostrophe in a single-quoted message (e.g. `git commit -m 'don'\\''t gh pr merge yet'`) still false-matches T3. The canonical case the fix claimed to close. |
| 2 | code-reviewer | `log-tier-events.sh` extended ANCHOR class | Bash array literal `result=(gh pr merge 88)` false-matches T3 because `(` was added to the leading-anchor class without distinguishing `$(` from bare `(`. New regression. |
| 3 | security-reviewer | All three hooks + settings traversal | **Working exploit reproduced**: symlink TOCTOU + attacker-controlled `journal.dir` from `.claude/settings.flow.local.json`. A hostile repo can redirect every T1/T2/T3 append to arbitrary writable paths. Arbitrary-file-append primitive on every Bash/Edit/Write tool call. |

### P2 — Should Fix (9)

- Heredoc bodies bypass quote-stripping (false positives from `cat <<EOF\ngit push origin\nEOF`)
- Shell comments (`# gh pr merge`) bypass quote-stripping
- Markdown comment-terminator injection: commit message containing `-->` corrupts journal entries
- Settings-file `journal.dir` accepts absolute paths and `..` traversal — needs validation
- Code duplication: same quote-strip + anchor logic in two files (drift risk)
- No test fixtures for parser logic — cycle-1→cycle-2 regression proves fragility
- Append interleave: two `echo >>` per hook invocation; not atomic across writes
- Malformed stdin causes `exit 5` (set -euo pipefail + jq failure) — breaks "graceful exit 0" contract
- Doc note about end-class punctuation is incomplete (`;`/`.` accepted as terminators is correct but undocumented)

### P3 — Consider (8)

DoS via unbounded journal growth, info disclosure of secret-like file paths, hook ordering undocumented, SKILL.md parser strategy text contradicts itself, narrow TOCTOU on `[ -f ]` then `>>`, settings stderr suppression, etc.

### Convergence assessment

| Cycle | Findings | Direction |
|---|---|---|
| 1 (during /flow:start) | 6 × P2 | — |
| 2 (this) | 3 × P1 + 9 × P2 + 8 × P3 | **DIVERGING** |

Per autonomous-workflow stop conditions: *"3+ verification failures on the same issue → Stop fixing forward. Return to EXPLORE — the problem is architectural."*

### Six-field escalation

**Situation**: Two cycles of fix-forward on text-pattern bash command parsing have produced more findings, not fewer. P1.1 is a regression of cycle-1 P2#1. P1.2 is a regression introduced by the cycle-1 fix-forward. P1.3 is a working security exploit. The architecture (regex+sed against `tool_input.command`) is unsound for the breadth of inputs it must handle.

**What I tried**: Quote-stripping with sed; tightening word-end and leading-anchor regex classes; extracting END/ANCHOR pattern constants for consistency. Each fix patched one case but missed adjacent ones. Also confirmed there's no fundamentally simpler signal source the harness exposes today (the `tool_input.command` text is what we have).

**Options**:
1. **Rearchitect parsing** — replace regex+sed with proper shell tokenization (e.g. `python3 -c \"import shlex; ...\"` to walk the command structure or refuse to classify). Substantial work; needs another spike. Ships T1+T2+T3 correctly.
2. **Scope-down to T1 only** — drop `log-tier-events.sh` entirely from this PR. Keep T1 tagging (which triggers on tool-name match, not command content — no parsing fragility). File a separate, more carefully designed issue for T2/T3. Unblocks Findings Ledger now (T1 only); Tier Summary stays blocked on the T2/T3 follow-up.
3. **Ship with security fix only + documented caveats** — fix the P1.3 security finding (symlink check + settings traversal validation). Accept P1.1 and P1.2 as documented best-effort limitations. Add a hook-header comment naming the known false-positive cases. Risk: spoofed entries can mislead `/flow:status` Tier Summary.
4. **Fix-forward again (cycle 3)** — try to patch each cycle-2 finding individually. Data suggests this won't converge.

**My recommendation**: Option 2 (scope down to T1 only). The text-parsing problem is real and worth a separate spike with proper tokenization, but it should not block the Findings Ledger work. T1 tagging is the trivially-safe part — tool name comes from the harness, not the command text. Splitting lets us ship the safe half today and engineer the parsing properly without time pressure.

**Time sensitivity**: Blocks `/flow:start 84` (Findings Ledger) and `/flow:start 85` (Tier Summary), which were waiting on this PR. Option 2 unblocks #84 immediately and re-blocks #85 only.

**Risk if wrong**: Option 1 risks scope creep and a third cycle of regressions on a more complex parser. Option 2 risks losing momentum on the Tier Summary feature. Option 3 ships a security fix but accepts known correctness issues. Option 4 risks a fourth review cycle.

I am pausing fix-forward and waiting for direction.

danielbentes:
## Closing — architectural rethink required

Per brainstorm decision: stopping work on this PR and deferring all journal tier tagging.

### Why

Cycle-1 self-review surfaced 6 P2 findings on the regex+sed command-parsing approach. Fix-forward addressed each, but cycle-2 review surfaced **3 P1 + 9 P2 + 8 P3** — including:

- A working security exploit (symlink TOCTOU + attacker-controlled `journal.dir` → arbitrary file append on every Bash/Edit/Write tool call, exploitable via a hostile repo's checked-in `.claude/settings.flow.local.json`)
- Two correctness regressions: apostrophes in commit messages (`git commit -m 'don'\''t gh pr merge yet'`) and bash array literals (`result=(gh pr merge 88)`) both false-match T3
- Multiple parsing gaps: heredocs, shell comments, escaped quotes, `-->` injection in journal entries

Per autonomous-workflow stop conditions (3+ verification failures on the same issue → return to EXPLORE — the problem is architectural), this signaled that text-pattern bash command parsing for high-blast-radius hooks is the wrong layer for the problem.

### Brainstorm outcome

Three approaches considered (Python shlex tokenization, wrapper aliases, scope-down to /flow commands only). Decision: **stop entirely** — the slide mockup's Tier Summary feature isn't load-bearing enough to justify either a parsing layer or a /flow-command-side helper migration right now.

### What this means downstream

- **Issue #84 (Findings Ledger)** — UNBLOCKED. The Findings Ledger reads `FLOW_RESOLUTION_CYCLE` markers from PR comments via `gh api`, not from journal tier events. It never actually depended on this prerequisite (the dependency I added when filing was over-cautious). Will be re-evaluated and likely proceed independently.
- **Issue #85 (Tier Summary)** — depends on this. With this PR closed, #85 is effectively deferred. If a future need surfaces, the brainstorm artifact is preserved in this PR's review thread for whoever picks it up.
- **The original journal hooks** (`log-file-changes.sh`, `log-commits.sh`) — UNCHANGED on main. No tier tags introduced. Backwards compat trivially preserved.

### Lessons

- Two cycles of fix-forward on a fundamentally fragile architecture is the autonomous-workflow's intended trip-wire. The signal worked.
- The original prerequisite issue I filed was scoped against the slide mockup. The brainstorm revealed the right question is "what tier semantics actually serve users?" not "how do we count every shell event?" — the "every shell event" framing is what made parsing necessary.
- Deferring is a valid outcome. Not every aspirational slide mockup needs to ship.

Branch deleted. Issue #90 will be closed as not-planned with a link back to this thread.

PR #117, opened 2026-05-23
fix(flow): merge-aware force-branch-delete in block-destructive hook

### LINKED ISSUE #116: fix(flow): block-destructive hook cannot delete squash-merged local branches
## Context

The flow plugin's `PreToolUse` hook `block-destructive.sh` blocks every force branch-delete and tells the user to use the safe-delete flag instead. But this marketplace merges PRs with **squash** (`settings.json` -> `merge.strategy: squash`). After a squash merge, the feature branch's commits never land on the default branch individually, so safe-delete refuses to delete them ("not fully merged") -- meaning force-delete is the only way to remove a squash-merged branch, and that's exactly what the hook blocks.

## Current State

Force branch-delete is blocked unconditionally. Routine post-merge local cleanup is impossible for the agent; merged feature branches accumulate (e.g. a branch left stranded after a squash merge can never be cleaned up automatically).

A second over-match was observed while filing this issue: the hook matches the force-delete pattern as a raw substring anywhere in the command, so even an `echo`/heredoc that merely *mentions* the command string is blocked. Considered out of scope here but noted for a follow-up.

## Objective

Allow force-deleting a local branch when its work is already contained in the default branch (including the squash-merge case), while continuing to block deletion of branches that carry genuinely unmerged work.

## Acceptance Criteria

- [ ] Force-delete is allowed when the branch is merged into the default branch via a normal/fast-forward merge.
- [ ] Force-delete is allowed when the branch was squash-merged (its combined change is patch-equivalent to a commit already in the default branch).
- [ ] Force-delete is still blocked when the branch has changes not present in the default branch.
- [ ] Force-deleting the default branch itself is always blocked.
- [ ] A multi-branch delete is allowed only if every target is merged; otherwise blocked.
- [ ] When merge status cannot be determined (git unavailable, no resolvable default branch, unresolvable ref), the command is blocked (fail-safe).
- [ ] All other destructive blocks remain unchanged: recursive force-remove (outside the safe-dir allowlist), hard reset, force clean, and discard-all-changes via checkout/restore.
- [ ] Tests in the flow harness cover the above and `bash plugins/flow/tests/run.sh` exits 0.

danielbentes:
## Summary

Closes #116. Makes the flow plugin's `block-destructive.sh` force-branch-delete rule **merge-aware** instead of a blanket block.

**Problem:** the hook blocked every force branch-delete and told the user to use safe-delete instead — but this repo squash-merges PRs, and safe-delete refuses to remove squash-merged branches (git doesn't see their commits on the default branch). So force-delete was the only option, and it was blocked → merged branches could never be cleaned up by the agent (which is exactly what stranded the `feature/issue-110` branch).

## Change

Force-delete is now allowed **only when every target branch is provably merged into the default branch**, and blocked otherwise (and on any uncertainty — fail-safe):

- **Merge detection:** `git merge-base --is-ancestor` for regular/fast-forward merges; a synthetic-commit + `git cherry` patch-equivalence check for squash merges (the branch's *combined* tree change vs the default). Checked against the local default **and** `origin/<default>`, so a branch merged on the remote is deletable even if the local default is behind.
- **All force-delete forms** are detected and routed through the same check — `-D`, combined clusters (`-Df`, `-fD`), and `--delete --force` / `-d --force` — not just bare `-D`.
- **Always blocked:** the default branch itself, unresolvable/nonexistent refs, unmerged work, missing git/default-branch, and force-deletes chained into compound commands (`&& …`).
- Branch names are passed to git as quoted args under `set -f`; no shell-injection or globbing surface.
- All other destructive blocks (`rm -rf`, `git reset --hard`, `git clean -f`, `git checkout -- .` / `git restore .`) are unchanged.

## Review

An independent review pass found 0 P1 (core safety property holds — no path allows deleting unmerged work) and 1 P2 (force-delete *equivalent* forms bypassed the trigger). That P2 and the P3 robustness items (origin-default merges, added test coverage) are fixed in this PR and re-verified.

## Test plan

- [x] New `tests/block-destructive-branch.test.sh` (24 assertions): regular-merged / squash-merged / unmerged / default-branch / multi-target / nonexistent / non-repo / safe-delete / equivalent forms / origin-default / compound-command / force-create — all via real scratch git repos driving the hook script.
- [x] `bash plugins/flow/tests/run.sh` → 809 pass / 0 fail.
- [x] Real-repo end-to-end: the stranded squash-merged `feature/issue-110` branch → allowed (via both `-D` and `--delete --force`); a genuinely unmerged branch → blocked.

## Bundled: flaky test-harness fix

This PR also includes a one-line fix to `tests/lib/assert.sh` (separate commit). `assert_match` piped the haystack into `grep -q`, which exits at the first match and closes the pipe — killing `printf` with SIGPIPE (141); under the CI runner's `set -o pipefail` the pipeline then reported non-zero despite a successful match, causing nondeterministic false failures on large haystacks (it surfaced here as a spurious `start-v3-integration` failure on Linux CI, never on macOS). Switched to a here-string so there is no pipeline. Pre-existing, suite-wide, and independent of the hook change — bundled because it was blocking this PR's CI. Verified: the CI `test` job, previously red on the race, is now green.

## Note

The live hook runs from the installed plugin cache, so this fix takes effect after the plugin updates from the marketplace (a release), not immediately in this session.

## Out of scope (noted for follow-up)

The hook matches its patterns as raw substrings, so a command that merely *mentions* a blocked pattern (e.g. an `echo`/heredoc) is also blocked, and `rm -f` of a single non-safe-dir file is treated like `rm -rf`. Both are pre-existing over-matches, unrelated to merge-awareness.

### DIFF
--- a/plugins/flow/hooks/scripts/block-destructive.sh
+++ b/plugins/flow/hooks/scripts/block-destructive.sh
@@ -33,31 +33,49 @@ if echo "$COMMAND" | grep -qE 'rm\s+(-[rfRF]+\s+)+|rm\s+(-[a-zA-Z]\s+)*-[a-zA-Z]
   fi
 fi
 
-# Force branch deletion (git branch -D): allowed ONLY when every target branch
-# is provably merged into the default branch. `git branch -d` (the safe form)
-# cannot delete squash-merged branches — git doesn't see their commits on the
-# default branch — so a blanket -D block makes squash-merged cleanup impossible.
-# This check distinguishes merged (safe to drop) from unmerged (would lose work)
-# and fails safe (blocks) on any uncertainty.
-if echo "$COMMAND" | grep -qE 'git[[:space:]]+branch[[:space:]]+-D([[:space:]]|$)'; then
+# Force branch deletion: allowed ONLY when every target branch is provably
+# merged into the default branch. `git branch -d` (the safe form) cannot delete
+# squash-merged branches — git doesn't see their commits on the default branch —
+# so a blanket force-delete block makes squash-merged cleanup impossible. This
+# check distinguishes merged (safe to drop) from unmerged (would lose work) and
+# fails safe (blocks) on any uncertainty.
+#
+# Force-delete is detected in every equivalent form: -D, a combined short
+# cluster containing D (-Df, -fD, -rD), or a delete flag (-d/--delete) together
+# with a force flag (-f/--force). Plain `git branch -d` (safe delete, no force)
+# is NOT matched and passes through.
+IS_FORCE_DELETE=0
+if printf '%s' "$COMMAND" | grep -qE 'git[[:space:]]+branch([[:space:]]|$)'; then
+  if printf '%s' "$COMMAND" | grep -qE '(^|[[:space:]])-[A-Za-z]*D[A-Za-z]*([[:space:]]|$)'; then
+    IS_FORCE_DELETE=1
+  elif printf '%s' "$COMMAND" | grep -qE '(^|[[:space:]])(--delete|-[A-Za-z]*d[A-Za-z]*)([[:space:]]|$)' \
+    && printf '%s' "$COMMAND" | grep -qE '(^|[[:space:]])(--force|-[A-Za-z]*f[A-Za-z]*)([[:space:]]|$)'; then
+    IS_FORCE_DELETE=1
+  fi
+fi
+
+if [ "$IS_FORCE_DELETE" = "1" ]; then
   set +e          # this block is fully conditional and always exits explicitly
   set -f          # no pathname expansion of parsed tokens
 
-  # Targets = non-flag tokens after the -D flag.
+  # Targets = every non-flag token that isn't `git`/`branch`. A force-delete
+  # chained into a compound command (`&& rm ...`, `; ...`, `| ...`) leaves
+  # junk tokens that fail the show-ref check below, so compounds always fall
+  # through to BLOCK rather than being allowed on a merged target.
   TARGETS=$(printf '%s' "$COMMAND" | awk '{
-    seen=0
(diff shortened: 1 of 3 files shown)


danielbentes (commented):
## Self-Review Summary

Performed self-review with fix-forward. An independent review pass examined this security hook with a correctness/safety bias; findings below were fixed in-PR and re-verified. No open findings on the current diff.

### Verdict Summary

| Overall | Criteria Passed | Criteria Failed | Human Review |
|---------|----------------|-----------------|--------------|
| PASS | 8 | 0 | 0 |

Core safety property holds: no input path allows force-deleting a branch with unmerged commits (verified empirically against regular / squash / partial-squash / disjoint branches). `bash plugins/flow/tests/run.sh` → 809 pass / 0 fail.

### Findings Found & Fixed
| # | Priority | Location | Issue | Fix Applied |
|---|----------|----------|-------|-------------|
| 1 | P2 | block-destructive.sh trigger | Force-delete equivalents (`--delete --force`, `-Df`, `-fD`, `-d -f`) bypassed the bare-`-D` trigger and fell through to allow | Broadened detection to all force-delete forms + generic target parsing; all routed through the merge check |
| 2 | P3 | block-destructive.sh default resolution | Merge checked only against the local default, so a branch merged on the remote but with a behind local default was false-blocked | Also accept merge into `origin/<default>` (safe direction only) |
| 3 | P3 | tests | Equivalent forms and the origin-default path were untested | Added scenarios for both (plus compound-command + force-create) |
| 4 | P3 | block-destructive.sh parsing | Compound commands parsed loosely | Confirmed fail-safe-to-block; junk tokens fail the ref check, so a chained tail is never greenlit (added a test) |

None — all findings fixed in-PR.

PR #121, opened 2026-05-25
fix(flow): copy $ARGUMENTS before bash parameter-expansion in command blocks

### LINKED ISSUE #120: flow: `${ARGUMENTS%% *}` in command !-blocks expands to empty (Claude Code substitutes bare $NAME only)
## Summary

The flow plugin's slash-command `!`-blocks use bash parameter expansion `${ARGUMENTS%% *}` to extract the first whitespace-separated token from the command argument. Claude Code's slash-command preprocessor substitutes bare `$ARGUMENTS` (and the bare `${ARGUMENTS}` form) into command markdown, but does **NOT** substitute bash parameter expansion syntax like `${ARGUMENTS%% *}`. The bash interpreter then sees an undefined `ARGUMENTS` variable and `${ARGUMENTS%% *}` expands to the empty string. The downstream `[ -n "$ISSUE_NUM" ]` guard short-circuits and the command silently degrades to its "no issue" path — even when the user passed an issue number.

## Reproduction

1. Install flow ≥3.1.0 (3.1.1 verified) into a project with `.claude/settings.flow.json` containing `flow.goals.requireGoalForStart: true`.
2. Invoke `/flow:start 69` (any digit-only issue number).
3. Observe Phase 0 PRE-FLIGHT output:

   ```
   ### Pre-Flight
   ISSUE_NUM=
   PREFLIGHT_ERRORS=2
   PREFLIGHT_WARNINGS=0
   PREFLIGHT_STATE=BLOCKED
   ```

4. Observe Phase 1 FlowGoal output:

   ```
   ### FlowGoal Auto-Creation
   FLOW_GOAL_STATE=skip
   ```

Both `ISSUE_NUM` is empty and `FLOW_GOAL_STATE=skip` — even though `requireGoalForStart=true` and the user passed `69`.

## Root cause

`plugins/flow/commands/start.md:55`:

```bash
ARG1="${ARGUMENTS%% *}"
case "$ARG1" in
  ''|*[!0-9]*) ISSUE_NUM="" ;;
  *) ISSUE_NUM="$ARG1" ;;
esac
```

Claude Code's `$ARGUMENTS` substitution operates on bare-`$NAME` (and `${NAME}`) literals — not bash parameter expansions. The bash code sees `${ARGUMENTS%% *}` as a literal; with `ARGUMENTS` not defined as a bash env var, the expansion produces `''`. ARG1 becomes empty; the case statement sets `ISSUE_NUM=""`; preflight fails and `FLOW_GOAL_STATE=skip` overrides the `requireGoalForStart=true` setting.

Confirmed empirically:

```console
$ env -i bash <<'BASH'
set +u
ARG1="${ARGUMENTS%% *}"
echo "ARG1='$ARG1'"
BASH
ARG1=''
```

Affected blocks in `commands/start.md` (3.1.1): line 55 (Phase 0 PRE-FLIGHT), line 370 (Phase 1 FlowGoal auto-creation), and the Phase 2 stranger-test manifest emit. The same `\${ARGUMENTS%% *}` pattern likely exists in other flow commands (`/flow:pr`, `/flow:address`, `/flow:review`, etc.) — please grep the plugin tree.

## Suggested fix

Two-line refactor that splits the Claude Code substitution from the bash parameter expansion:

```bash
# Before
ARG1="${ARGUMENTS%% *}"

# After
_RAW="$ARGUMENTS"        # Claude Code substitutes bare $ARGUMENTS
ARG1="${_RAW%% *}"       # bash now operates on a real env var
```

`$ARGUMENTS` (bare) IS substituted by Claude Code's preprocessor — the title line `# Start Work on Issue #\$ARGUMENTS` works correctly. The fix moves the substitution out of the parameter-expansion context where Claude Code's matcher doesn't reach.

Alternatively, document the limitation in the plugin's contributor guide ("never use bash parameter expansion on `\$ARGUMENTS` inside a `!`-block"). The current pattern silently degrades; any guard / fix would have surfaced this years ago.

## Impact

- `/flow:start <N>` with `flow.goals.requireGoalForStart=true` skips goal creation — the gate is documented as load-bearing but never fires.
- `/flow:pr` then reports `STATE=missing, GATE=block` and asks the user how to proceed, contradicting the v3 lifecycle promise.
- `/flow:start` also reports `PREFLIGHT_STATE=BLOCKED` — users learn to ignore Phase 0 PRE-FLIGHT, defeating its purpose.

## Environment

- macOS Darwin 25.5.0
- Claude Code: latest (Opus 4.7)
- flow plugin: 3.1.1 (`/Users/danielbentes/.claude/plugins/cache/synapti-marketplace/flow/3.1.1/`)
- bash: macOS default `/bin/bash` (verified with `env -i bash`)

## Workaround

Manually invoke `/flow:goal create issue <N>` after `/flow:start` returns, OR proceed without a FlowGoal and journal the gate-override (the path I took for [memomee/memomee #69](https://github.com/synaptiai/memomee/issues/69)).

danielbentes:
## FlowGoal Status

- **Goal**: `issue-120` (`.flow/goals/issue-120.goal.yaml`)
- **Lifecycle**: `achieved` — all 4 acceptance criteria pass deterministically
- **Evidence**: per-criterion FlowEvidence sidecars under `.flow/runs/20260525T1810Z-issue-120-eval/evidence/` (local; `.flow/runs/` is gitignored)

## Summary

Claude Code's slash-command preprocessor textually substitutes only the bare `$ARGUMENTS` and `${ARGUMENTS}` tokens into command markdown — it does **not** substitute bash parameter-expansion forms like `${ARGUMENTS%% *}` or `${ARGUMENTS:-}`. Those left bash to expand an undefined `ARGUMENTS` var to the empty string, so first-token extraction yielded `""` and downstream gates (`PREFLIGHT_STATE`, `FLOW_GOAL_STATE`) silently degraded to their no-arg paths even when an argument was passed.

This bug self-demonstrated while starting work on the issue: `/flow:start` rendered `ISSUE_NUM=` (empty), `PREFLIGHT_STATE=BLOCKED`, `RUN_ID=...issue-nonum`.

### Why it was never caught

- The onboarding fixture ran the gate block as `ARGUMENTS="42" bash -c "$BLOCK"` — injecting `ARGUMENTS` as a real env var, so `${ARGUMENTS%% *}` expanded correctly *in the test*. Production never sets that env var.
- The lint (`command-frontmatter` Test 3) only checked that `$ARGUMENTS` was *quoted* and explicitly blessed `"${ARGUMENTS%% *}"` as the "case-validated form" — it never modeled the text-only substitution.

Both stayed green while production was broken.

## Fix

Replace the parameter-expansion-on-`$ARGUMENTS` pattern with a bare-substitution-first copy:

```bash
_RAW="$ARGUMENTS"   # Claude Code substitutes the bare arg token
ARG1="${_RAW%% *}"  # bash parameter-expansion now hits a real var
```

Applied to **16 sites across 8 command files**: `start.md` (5), `merge.md` (3), `address.md`/`review.md` (2 each), `resolve.md`/`design.md`/`brainstorm.md` (1 each), plus `resume.md`'s `RUN_ID="${ARGUMENTS:-}"` (same bug class in a ```bash fence, caught during the sweep). `workflow.md`'s `${ARGUMENTS}` (bare braces) is correctly left alone — that form *is* substituted.

## Regression guards

- **`command-frontmatter` Test 6** (new): rejects bash parameter-expansion on `$ARGUMENTS` in `!`/`bash`/`sh` fences, with a comment-strip anchored to word boundaries so it covers every operator (including the `#`/`##` prefix forms) and a self-guard asserting that coverage with no false positives.
- **`flow-start-onboarding` fixture** (rewritten): a `cc_substitute` helper models Claude Code's bare-only text substitution instead of env-var injection. A regression assertion proves the old form degrades to `skip` while the fixed `_RAW` form produces `create` under the identical substitution model — so a revert fails loudly.

## Acceptance criteria (derived — issue had none)

| # | Criterion | Verification | Result |
|---|-----------|--------------|--------|
| AC1 | No `${ARGUMENTS<op>}` param-expansion in any command executable block | `! grep -rqE '\$\{ARGUMENTS[%#/:^,@+-]' plugins/flow/commands/` | exit 0 |
| AC2 | Onboarding fixture models CC substitution and proves bug + fix | `run.sh flow-start-onboarding.test.sh` | 73 pass |
| AC3 | Static lint rejects the broken form, self-guards operator coverage | `run.sh command-frontmatter.test.sh` | 10 pass |
| AC4 | Full flow test suite passes | `run.sh` | 840 pass |

## Review

Independent code-review pass (6-facet): **0 P1, 1 P2, 1 P3 — both fixed in-PR**.
- **P2 (F1)**: the lint's comment-strip would have truncated `${ARGUMENTS#foo}`/`${ARGUMENTS##*/}` at the `#`, silently dropping 2 of the 9 operators it advertised. Fixed with a word-boundary strip + synthetic self-guard.
- **P3 (F2)**: clarified the `sh`-fence coverage comment (deliberate forward-coverage).

Closes #120

### DIFF
--- a/.decisions/issue-120.md
+++ b/.decisions/issue-120.md
@@ -12,6 +12,12 @@ artifacts:
   result: achieved
   evidence_bundle: .flow/runs/20260525T1810Z-issue-120-eval
   failures: none
+- type: review-cycle
+  captured_at: '2026-05-25T16:35:50Z'
+  cycle: 1
+  path: B
+  findings_count: 2
+  pr: 121
 ---
 # Issue #120 — `${ARGUMENTS%% *}` in command !-blocks expands to empty
 
@@ -142,3 +148,5 @@ fixture fix + lint guard).
 <!-- auto-log: 2026-05-25 18:34 Edit /Users/danielbentes/synapti-marketplace/.gitignore -->
 
 <!-- auto-log: 2026-05-25 18:34 commit "chore(flow): add achieved FlowGoal contract for issue 120" -->
+
+<!-- auto-log: 2026-05-25 18:36 commit "chore(flow): link FlowGoal issue-120 to PR #121" -->
--- a/.decisions/issue-120.md
+++ b/.decisions/issue-120.md
@@ -18,6 +18,12 @@ artifacts:
   path: B
   findings_count: 2
   pr: 121
+- type: review-cycle
+  captured_at: '2026-05-25T16:44:49Z'
+  cycle: 2
+  path: B
+  findings_count: 1
+  pr: 121
 ---
 # Issue #120 — `${ARGUMENTS%% *}` in command !-blocks expands to empty
 
@@ -150,3 +156,13 @@ fixture fix + lint guard).
 <!-- auto-log: 2026-05-25 18:34 commit "chore(flow): add achieved FlowGoal contract for issue 120" -->
 
 <!-- auto-log: 2026-05-25 18:36 commit "chore(flow): link FlowGoal issue-120 to PR #121" -->
+
+<!-- auto-log: 2026-05-25 18:42 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/tests/command-frontmatter.test.sh -->
+
+<!-- auto-log: 2026-05-25 18:43 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/tests/command-frontmatter.test.sh -->
+
+<!-- auto-log: 2026-05-25 18:43 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/tests/command-frontmatter.test.sh -->
+
+<!-- auto-log: 2026-05-25 18:43 Edit /Users/danielbentes/synapti-marketplace/plugins/flow/tests/command-frontmatter.test.sh -->
+
+<!-- auto-log: 2026-05-25 18:44 commit "test(flow): broaden $ARGUMENTS lint to length/indirection/array forms" -->
--- a/.decisions/issue-120.md
+++ b/.decisions/issue-120.md
@@ -1,0 +1,119 @@
+# Issue #120 — `${ARGUMENTS%% *}` in command !-blocks expands to empty
+
+**Branch**: `fix/issue-120-arguments-param-expansion`
+**Type**: bug (flow plugin self-fix)
+
+## Problem
+
+Claude Code's slash-command preprocessor substitutes bare `$ARGUMENTS` and
+`${ARGUMENTS}` (no operator), but NOT bash parameter expansions like
+`${ARGUMENTS%% *}`. The bash interpreter then sees an undefined `ARGUMENTS`
+variable and the expansion yields the empty string, so `ISSUE_NUM` ends up
+empty and downstream gates (`PREFLIGHT_STATE`, `FLOW_GOAL_STATE`) silently
+degrade to their no-issue paths even when the user passed a number.
+
+Self-demonstrated this session: invoking `/flow:start <120 URL>` rendered
+`ISSUE_NUM=` (empty), `PREFLIGHT_STATE=BLOCKED`, `RUN_ID=...issue-nonum`.
+
+## Why it was never caught
+
+The unit fixture `flow-start-onboarding.test.sh` ran the gate block as
+`ARGUMENTS="42" bash -c "$BLOCK"` — injecting `ARGUMENTS` as a real bash env
+var, so `${ARGUMENTS%% *}` expanded correctly *in the test*. The lint
+`command-frontmatter.test.sh` Test 3 only checked that `$ARGUMENTS` was
+*quoted* (word-split safety) and explicitly blessed `"${ARGUMENTS%% *}"` as
+the "case-validated form". Neither test modeled Claude Code's text-only
+substitution, so both stayed green while production was broken.
+
+## Specification
+
+### Non-goals
+- Not changing the first-token extraction *semantics* (still strip trailing
+  prose; still reject non-digit input where it was rejected before).
+- Not the documentation-only alternative from the issue — the code refactor
+  is definitive; docs alone leave the silent-degradation behavior.
+- Not touching bare `$ARGUMENTS` uses in prose / agent-prompt strings (those
+  are substituted correctly).
+
+### Failure modes
+- Lint over-matches and flags legitimate `${SOMEVAR%% *}` on non-ARGUMENTS
+  vars → anchor strictly to `${ARGUMENTS<op>`.
+- Refactor introduces an unquoted `$ARGUMENTS` → existing Test 3 still guards
+  quoting; keep it.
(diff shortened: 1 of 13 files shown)


danielbentes (commented):
## 🔍 Self-review — cycle 2 (full post-fix diff)

Re-reviewed the **current** diff (which grew since the `/flow:start` review): the F1 fix, the resume.md comment reword, the FlowGoal artifact, and `.gitignore`. Two independent passes — a 6-facet code review and a convention check.

### Findings this cycle: P1: 0, P2: 0, P3: 1 — all resolved in-PR

| # | Priority | Category | Location | Status |
|---|----------|----------|----------|--------|
| F1 | P3 | correctness | `plugins/flow/tests/command-frontmatter.test.sh` (Test 6) | ✅ Resolved |

**F1** — the regression lint's operator class caught trailing-operator forms (`${ARGUMENTS%% *}`, `${ARGUMENTS:-}`, …) but missed three other shapes Claude Code *also* leaves unsubstituted: length `${#ARGUMENTS}`, indirection `${!ARGUMENTS}`, and array subscript `${ARGUMENTS[0]}`. Silent-degradation vectors the guard couldn't see. **Fixed**: centralized the pattern in a single `_ARGEXP_RE` (one source of truth for the scan + its self-guard) and extended the self-guard to assert all three new forms are flagged. (commit `test(flow): broaden $ARGUMENTS lint…`)

### Convention check: clean
- 6 commits, all `<type>(<scope>): <subject>`; no Claude attribution; branch naming correct; no issue/finding refs in code comments or test prose.
- One PR-meta nit: missing label → **added `bug`**.

### Verified-correct (not findings)
- The earlier-fixed **F1/F2 did not regress**.
- Word-boundary comment-strip `s/(^|[[:space:]])#.*$/\1/` is correct (catches `#`/`##` operators, no doc-comment false positives); self-guard is non-tautological.
- `cc_substitute` faithfully models CC bare-only substitution; regression arm genuinely fails on the old form.
- All 8 command-file sites preserve first-token-extraction semantics; `.flow/goals/*.lock` ignore matches the real lock filename while keeping the goal YAML committable.

### Gate
- Full suite: **840 pass / 0 fail** · Lint: **10/10** · FlowGoal `issue-120`: **achieved** (AC1–AC4 deterministic)

Ready to merge.

danielbentes:
## ✅ Resolution — cycle 2

The single P3 finding from review cycle 2 was resolved in-PR via self-review fix-forward (no separate address pass was needed).

| Finding | Priority | Resolution |
|---------|----------|------------|
| F1 | P3 | **Resolved** in commit `fa857a0` — broadened the `$ARGUMENTS` lint to also flag length (`${#ARGUMENTS}`), indirection (`${!ARGUMENTS}`), and array-subscript (`${ARGUMENTS[0]}`) forms; centralized the pattern in `_ARGEXP_RE` and extended the self-guard to assert all three. |

Verification: full suite 840/840, lint 10/10 (self-guard covers 12 broken forms), FlowGoal `issue-120` achieved.

PR #131, opened 2026-07-25
feat(dossier): evidence-first documentation plugin with post-merge automation

danielbentes:
Turns the universal project documentation system — previously a ~5,000-word prompt you paste into a chat — into a marketplace plugin, and gives it a maintenance path.

The prompt had two problems. It could not be selectively loaded, so every run carried all per-file content requirements in context. And it had no way to stay current: the source document itself lists the change events that should trigger a refresh (architecture, interfaces, data handling, security controls, public claims, releases), every one of which is what a merged PR represents, but nothing connected them. The package went stale the day it was written and then *looked* current — the failure it exists to prevent.

## What ships

`plugins/dossier` — 9 skills, 9 commands, 6 agents, 20 references, 23 package templates, 14 scripts, 4 hooks, 15 test files. Marketplace `4.6.2` → `4.7.0` under a new `documentation` category.

The generated package is 23 files across 8 numbered directories, produced for every project. Content adapts to project type; structure never does, because a package whose shape varies cannot be diffed, audited, or compared across engagements — and "we dropped that file, it did not apply" is indistinguishable from "we ran out of time" six months later.

## What makes it more than a docs generator

**Every material claim carries an evidence ID and a claim state.** Verified / corroborated / reported / inferred / unknown / not-applicable. Only the first two may appear unqualified in a public document. An untraceable claim is written as `Unknown` rather than as confident prose.

**Three verification passes that genuinely cannot see each other.** Independence is architectural, not instructed: separate dispatches, `memory: none`, a skill firewall keeping reconciliation logic out of every verifier's context, single-message dispatch, and per-pass model configuration. `agent-independence.test.sh` enforces all five — a verifier that quietly gained the merge logic would still emit a plausible findings table, and nothing else would catch it.

**A gate that cannot be faked.** Sixteen conjunctive conditions: nine mechanical, seven requiring a model to read the package. `dossier-gate.sh` structurally refuses to emit `PASS` without a scorer verdict, because link-checking and header parsing certifying a package whose security claims nobody read is the exact theater this exists to prevent. Failing is decidable from a subset; passing is not.

**The findings table is published before any repair**, so the audit trail survives the fixes.

## The post-merge job

`/dossier:setup` renders a workflow into the consuming repository that regenerates affected documents after a PR merges and opens a documentation PR.

Three jobs split by privilege. `policy` decides and builds evidence with no agent. `refresh` runs the agent with `contents: read`, no write token, and no persisted git credentials — its entire output is a patch artifact. `publish` holds the write token and runs no agent code. A fully prompt-injected agent's maximum achievable outcome is a patch the privileged job rejects against a path allowlist.

Refresh is range-based against a durable cursor rather than the triggering PR's diff, which is what makes concurrent merges, dropped runs, retries, and the weekly sweep compose safely — a lost run is never a lost change, the next run's range just widens. Four loop guards are evaluated before a runner is allocated. Nothing force-pushes. Missing credentials fail the run loudly rather than skipping it, because stale documentation must never masquerade as current.

## Review notes

Three defects were found and fixed after the first pass, each now pinned by a regression test:

- The package is 8 directories, not 7. The wrong count reached the `doc-package-contract` Iron Law, so the plugin was asserting something its own scaffold contradicted.
- Contract volume was an uncounted estimate (~350); the real figure is 527 normative bullets.
- The pass C agent simulated all seven reader personas while its own protocol reference assigned one to pass A — an agent and its reference disagreeing about who audits what, which is precisely the cross-document contradiction this format exists to surface.

A fourth gap was found in the test suite itself: nothing cross-checked each contract's *declared* header style against the template's *actual* one. A contract declaring `internal-v1` for a public document would have passed while instructing a drafter to leak internal fields into a published file. Confirmed by mutation before and after the fix.

Two bugs were found in my own scripts by their tests. `jq`'s `//` is falsy-triggered, so `.dossier.ci.enabled // null` returned `null` for an explicit `false` — a feature a user disabled would silently re-enable itself on the next cascade layer. And the disclosure scanner was echoing the credentials it detects into its own findings output, destined for a CI log.

## Verification

1190 assertions across 15 test files; flow's 1022 unaffected. Mechanically verified end to end: scaffold creates 23 files and is idempotent (a re-run left a hand-added marker intact), ledger lint and disclosure scan behave correctly on fixtures, the gate reports `FAIL` with all seven judgment conditions `INCONCLUSIVE` on an undrafted package, and the workflow template renders to valid YAML with the correct job graph and permission split.

Not verified, and needing a human: the slash commands end to end in a live session, and a real merge triggering the workflow in a repo with the API key secret set.

## Round 4 — an external review of the plugin

A fourth review pass found five containment defects in code the earlier rounds had
passed over, all of them in the three files no prior cycle had touched:

- The output-root hook admitted a path traversal. Its allow-case is a `case` glob,
  and a glob `*` matches `/`, so `docs/dossier/../../.github/workflows/evil.yml`
  starts with the allowed prefix and lands outside it. The comment three lines below
  asserted the opposite behaviour.
- All three security hooks no-opped when `jq` was missing, turning a containment
  guarantee off silently. They now fail closed.
- The action ceiling was defeated by one layer of indirection — `/usr/bin/curl`,
  `env curl`, `bash -c "npm test"`, `eval "…"`. Wrappers are now detected and
  rescanned, while grepping for a command name still passes.
- The disclosure hook had no `internal-path` class, although the policy reference
  names it and says this hook enforces it alongside the batch scanner.
- The patch validator checked path strings but not file mode, so a symlink under an
  allowlisted path escaped by reference.

A sixth was found by a test written during the cycle rather than by a reader: the
validator also accepted `docs/dossier/../../.env`, which matches the allowlist's
compiled `^docs/dossier/.*$`.

The sharpest finding was not a defect in the code. **The regression test written to
pin the package's directory count reported a pass on the file that still carried the
defect** — its pattern required a space and the text was hyphenated. A green
assertion certifying the bug it was written to prevent is worse than no assertion,
and it had been green for three rounds. The count and the gate's condition split are
now derived from the artifacts themselves, so the expectation moves when they do.

Five more of the same class: the plugin README understated the gate by a condition
and its mechanical half by one; this README said four gate conditions fail when the
package has said two since round 3; two references cited `AS-` and `OQ-` registers
that exist nowhere; G17 sat out of sequence. One figure was mislabelled rather than
stale — `git rev-list --count HEAD` counts the assessed branch, and five documents
printed it as the commit count of `main`.

Test coverage grew where the gaps were load-bearing: the write-allowlist control had
no functional test at all and now has fourteen, almost all rejection cases; the
`LEAKED-FIELD` disclosure check had none; and the fifth independence mechanism was
unasserted while the README claimed a test enforced all five.

## Known limitations, stated in the README rather than discovered later

- `plugin_marketplaces` accepts no ref, so CI always runs the skill text from marketplace `main` even when helper scripts are pinned. `ci.instructionSource: vendored` is the complete fix.
- The circuit breaker bounds runaway loops, not cost per run. GitHub Actions has no spend cap for third-party API calls.
- Secret-scan patterns miss novel formats. The agent has no network egress and the docs PR is human-reviewed; those are the backstops.
- A plugin cannot guarantee a *different model* for verification. In-plugin passes give independent context; `/dossier:audit --external` is the genuine cross-model path. Which tier was used is recorded in the verification report.

## After merging, adopting repos still must

Add `ANTHROPIC_API_KEY` (or `CLAUDE_CODE_OAUTH_TOKEN`) as a secret, and enable **Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests"**. The latter is the most common setup failure: without it the branch pushes cleanly and PR creation is refused, so you get a successful push and no PR. On org-owned repos the org-level setting must be enabled first — until then the repo checkbox is greyed out and the API call returns 200 with no effect. `/dossier:setup` preflights both and prints the exact commands.


danielbentes (commented):
## Self-Review: PR #131 — cycle 1

Scope as requested: **does the plugin follow current Claude API / Claude Code conventions?** Reviewed against the bundled `claude-api` skill (model catalog, Agent-tool dispatch semantics, skill-authoring conventions, Opus 5 prompt-authoring guidance).

### Findings: P1: 2, P2: 2, P3: 0 — all fix-forwarded in this PR

### P1 — Critical

| Finding | Suggested Fix |
|---|---|
| **F1** · P1 · correctness · `plugins/dossier/commands/audit.md`<br>Phase 2 instructed the agent to "apply `verification.passModels` per pass". The shipped default for all three passes is `"inherit"`. The Agent tool's per-invocation `model` override accepts only `sonnet`/`opus`/`haiku`/`fable` — `inherit` is valid in agent *frontmatter* (meaning "use the session model") but is **not** a valid override value. A literal reading on the out-of-the-box config raises an InputValidationError at dispatch, breaking `/dossier:audit` at the one moment it exists for. | **Applied.** Added the value table: `inherit` → omit the argument entirely; `sonnet\|opus\|haiku\|fable` → pass as override; anything else → refuse to dispatch and report a config error. States explicitly that omitting and passing `"inherit"` are not the same thing. Flow documents this identical trap for `AGENT_TEAM_MODEL`, so it is a known Claude Code hazard, not a hypothetical. |
| **F2** · P1 · completeness · `plugins/dossier/templates/external-audit-prompt.md`<br>The file did not exist. `commands/audit.md` and `references/independent-audit-protocol.md` both render it for `/dossier:audit --external`, and the README names that flag as the mitigation for the plugin's own stated limitation that it cannot guarantee a different verification model. The only documented path to genuine cross-model independence was dead. | **Applied.** Written as a self-contained prompt — it assumes no access to the plugin, since it is handed to a different model or a human. Carries the claim-state vocabulary, the nine audit steps, the ten falsification traces, the scorecard, and the gate conditions inline. |

### P2 — Should fix

| Finding | Suggested Fix |
|---|---|
| **F3** · P2 · completeness · `plugins/dossier/commands/setup.md`<br>`ci.instructionSource: vendored` appeared in `schema.json`, README, CHANGELOG, the workflow comments, and as an offered setup option described as "the complete fix" for the `plugin_marketplaces` no-ref limitation and the answer for private marketplaces — while nothing anywhere wrote a vendored file. | **Applied.** Setup now emits `.claude/dossier/refresh-instructions.md`, drops `plugin_marketplaces`/`plugins` from the rendered workflow, and points `prompt` at the vendored file. The re-run-setup-to-update trade is stated where the user makes the choice. |
| **F4** · P2 · correctness · `plugins/dossier/commands/baseline.md` (+6)<br>`--max-drafters` was advertised in `argument-hint` and handled nowhere. Generalizing the check found six more: `audit/gate/status --path`, `claim --audience`, `init --mode`, `setup --ci`. A flag that parses and then does nothing is worse than an absent one — the user believes they set a bound that was never applied. | **Applied.** All seven handled. `--max-drafters` now defaults to **6**: current models delegate more readily than their predecessors, and baseline fans out one drafter per document across 21 documents, so an uncapped wave multiplies cost without improving any single document. |

### Conventions verified clean

- **Model IDs** — the only literal IDs in the plugin are `claude-opus-5` in reference examples. Current per the model catalog; no retired IDs, no date-suffixed aliases.
- **Agent frontmatter** — all 6 agents use `model: inherit`, which is correct for that surface.
- **`--model` emission in CI** — `bin/dossier-policy.sh` initialises `MODEL_ARG=""` and only sets `--model $MODEL` when non-empty, so an unset `ci.agent.model` never emits a malformed `--model ""`.
- **Skill descriptions** — all 9 carry real trigger conditions (`Use when …`), not bare capability statements. This is the property that governs on-demand loading.
- **Progressive disclosure** — SKILL.md bodies are 120–137 lines; the ~500 per-document requirements live in `references/package-contract-*.md`, one per directory, so a drafter loads exactly one.
- **Self-verification scaffolding** — checked for the over-verification pattern. The plugin's verification is a produced artifact under an audit protocol, not "double-check your work" instructions in an agent prompt. No change needed.

### Regression tests

Both new checks are deliberately general rather than special-cased — each found more than the instance that prompted it:

- Every `templates/` path named **in prose** must exist. Markdown-link checking never saw F2, because the file was referenced in a sentence rather than a link.
- Every flag in an `argument-hint` must appear in its command body. This is what turned one dead flag into seven.

### Quality gates

```
dossier  TOTAL pass=1030 fail=0   (+29 from the new checks)
flow     TOTAL pass=1022 fail=0   (unaffected)
CodeQL   pass
```

### Note on review path

Two paired reviewers were dispatched but had not returned when this was posted. The findings above are from single-session analysis, so this cycle uses the 5-field marker. If the paired reviewers surface anything new it will land as cycle 2 rather than being folded in silently here.

danielbentes:
## Resolution — cycle 1

All four findings from the cycle-1 self-review were fix-forwarded in this PR (`a1ed0b2`). Nothing deferred, nothing escalated.

| Finding | Priority | Resolution |
|---|---|---|
| F1 | P1 | `commands/audit.md` now carries the `passModels` value table; `inherit` omits the dispatch override rather than passing an invalid enum value |
| F2 | P1 | `templates/external-audit-prompt.md` written — self-contained, assumes no plugin access |
| F3 | P2 | `commands/setup.md` implements the vendored instruction source and states its update trade-off |
| F4 | P2 | All seven dead flags handled; `--max-drafters` defaults to 6 |

**Tests added for each fix** — both generalized so they catch the next instance rather than these:

- `references-integrity.test.sh` — every `templates/` path named in prose must exist. Caught F2's class; markdown-link checking could not see it.
- `command-frontmatter.test.sh` — every `argument-hint` flag must be handled in the body. Turned F4 from one dead flag into seven.

Gates after fixes: dossier **1030 pass / 0 fail** (+29), flow **1022 pass / 0 fail**.

danielbentes (commented):
## Self-Review: PR #131 — cycle 2 (paired reviewers)

The two paired reviewers dispatched during cycle 1 returned after that cycle was posted. Both corroborated the cycle-1 findings independently, and surfaced three the first pass missed — one of which showed a cycle-1 fix was incomplete.

### Findings: P1: 3, P2: 2 — all fix-forwarded in `d0fa737`

| Finding | Suggested Fix |
|---|---|
| **F5** · P1 · correctness · 2/2 corroborated · `bin/cascade-resolve.sh`, `bin/dossier-resolve-config.sh`<br>The cascade treated an explicit empty string as absence and fell through to a lower-precedence layer. Same defect class as the boolean `//` bug fixed one commit earlier, never extended to strings — where empty is equally meaningful (`project.name: ""` = deliberately unresolved, `ci.agent.model: ""` = inherit the action default). **Reproduced:** a prior client's name in user-global settings beat a new project that explicitly cleared it, and would propagate into every document header, the terminology register, and from there into public documents — the exact failure the schema comment says the empty default prevents. | **Applied.** Absence is decided by `jq -e '(path) != null'` rather than a shell emptiness test; the env layer uses `${VAR+x}` so an exported empty value is honoured as the deliberate clear it is. Verified across all five states: explicit empty, explicit false, absent, env-empty, env-set. |
| **F6** · P1 · correctness · `templates/ci/dossier-docs-refresh.yml:410`<br>The CI allowlist contained neither `Agent` nor `Skill`, while `commands/refresh.md` dispatches a collector, one drafter per affected document, and loads four skills — 5 invocation sites. In a headless run with no TTY to approve them, every one is unreachable, so the refresh silently degraded to flat inline drafting and lost the per-document context isolation the design depends on. No signal that it happened. | **Applied.** Both allowed, with the reasoning inline. They do not widen the blast radius: subagents inherit the job's own permissions (`contents: read`, no write token, no persisted credentials), and containment is enforced at the patch layer by the write allowlist in the publish job — not by the tool list. |
| **F7** · P1 · completeness · `schema.json` — **cycle-1 fix was incomplete**<br>F1 documented the `inherit` → omit rule in `audit.md` but left `passModels.*` an unconstrained string. A user following the plugin's *own* example table in `references/independent-audit-protocol.md:59` would set `claude-opus-5` — a full model ID, also invalid at the Agent-tool surface. Both the default value and the value the docs imply were wrong. | **Applied.** `enum: ["inherit","sonnet","opus","haiku","fable"]`, so an invalid value now fails at config-validation time instead of at dispatch. |
| **F8** · P2 · completeness · `references/release-gate-conditions.md`, `bin/dossier-gate.sh`<br>The reviewer-pass independence disclosure is called *required* by the protocol and was enforced by no gate condition. A package could satisfy every condition while omitting the table that says which passes ran on which models — the one sentence that tells a reader how much the verification is worth, since three passes sharing a model decorrelate lenses but not model-level blind spots. | **Applied.** Added as **G17**, mechanical, appended rather than renumbered. Count propagated through the reference, the skill, the command, the scorer agent, and the script. Gate now reports 17 conditions. |
| **F9** · P2 · correctness · `schema.json` `ci.agent.model`<br>Unconstrained free string interpolated into the workflow's `claude_args` block — whitespace would split into stray arguments for the action's parser. | **Applied.** No-whitespace pattern. This is the CLI surface, so a full model ID remains legitimate here — unlike `passModels`, which is the Agent-tool surface. The distinction is now stated in both descriptions. |

### Reviewer disagreement

None. Where both reviewers covered the same ground they reached the same conclusion, including on all four cycle-1 findings. The skeptic additionally flagged the evidence-collector batch fan-out as having no numeric cap (P3); I judged that one **not worth acting on** — batch count is bounded by source-inventory size rather than by open-ended delegation, unlike the drafter wave which fans out per document. Recording the judgment rather than silently dropping it.

### Regression tests

One per defect, all generalized:

- **cascade** — an explicit `""` at a higher layer must not fall through to a stale lower one
- **workflow** — `Agent` and `Skill` must appear in the allowlist, because `refresh.md` dispatches them. (The first version of this test passed spuriously: a bare `grep 'allowedTools'` matches the substring inside `--disallowedTools` in the header comment. Anchored on `--allowedTools "`.)
- **gate** — G17 must be evaluated

### Quality gates

```
dossier  TOTAL pass=1034 fail=0   (+4 from cycle 2, +33 total across both cycles)
flow     TOTAL pass=1022 fail=0
```

danielbentes:
## Resolution — cycle 2

All five cycle-2 findings fix-forwarded in `d0fa737`. Nothing deferred, nothing escalated.

| Finding | Priority | Resolution |
|---|---|---|
| F5 | P1 | Cascade decides absence via jq, not shell emptiness; env layer tests set-ness. All five states verified |
| F6 | P1 | `Agent` and `Skill` added to the CI allowlist with the containment reasoning inline |
| F7 | P1 | `passModels.*` constrained to an enum — completes the incomplete cycle-1 fix |
| F8 | P2 | G17 added, mechanical, count propagated across six files |
| F9 | P2 | `ci.agent.model` pattern-constrained against whitespace splitting |

One reviewer finding was assessed and **not** actioned, recorded rather than dropped: the evidence-collector batch fan-out has no numeric cap, but batch count is bounded by source-inventory size rather than open-ended delegation — unlike the drafter wave, which fans out per document and does have a cap.

Gates: dossier **1034 pass / 0 fail**, flow **1022 pass / 0 fail**.

danielbentes (commented):
## Self-Review: PR #131 — cycle 3

This cycle reviews the dogfood run itself: pointing the plugin at this repository, and what that surfaced. Method diversity was the substitute for reviewer independence — one pass checked referential integrity mechanically, one re-derived every headline count from the filesystem **without consulting the ledger**, one compared repeated facts across all 23 documents. That is recorded as a limitation, not a strength; see F-05.

### Findings: P1: 6, P2: 3, P3: 1 — 9 fix-forwarded, 1 open by design

Six are defects **in the plugin**. Its 1034 assertions caught none of them, because each is a disagreement between two artifacts no single test compared.

| Finding | Suggested Fix |
|---|---|
| **F-06** · P1 · correctness · `bin/dossier-gate.sh`, `templates/package/07-verification/…`<br>Gate condition G17 greps the verification report for `## Reviewer-pass independence method` and a `Model diversity:` line. The shipped template carried neither — its heading was `## Independence method`, and it never prompted for the line. G17 was added in the previous cycle without touching the artifact it grades. | **Applied.** Template renamed to match the gate and the contract reference; the prompt added. The test reads G17's expected heading **out of the gate script** rather than restating it, so renaming either side alone fails. |
| **F-07** · P1 · correctness · `bin/dossier-ledger-lint.sh`<br>Two defects in one check. The linter compared `Source ref` against the filesystem verbatim while `references/evidence-ledger-schema.md` writes every one of its own examples as a code span — so the documented form failed the lint, 53 of 82 first-run errors. Then the multi-locator loop the fix needed ran in a pipe, putting `emit` in a subshell: **the linter printed findings under `LEDGER_ERRORS=0`.** | **Applied.** Code spans stripped before classification; every span in a cell validated; loop de-piped. A caller reading only the count no longer sees a clean ledger while findings print. Added `derived:` as a locator form — a row grounded in other rows was always allowed by the method, and the schema table had no form for it. |
| **F-08** · P1 · correctness · `bin/dossier-claim-scan.sh`<br>Approved wordings were lowercased but not normalized, while document sentences were fully normalized. Markdown survived on one side only, so **any claim containing a code span could never match its own approved row** — the registration check could not pass for a realistic claim. Separately, sentences split on a bare `.`, cutting inside `SKILL.md`, `plugin.json`, and version numbers. | **Applied.** Both sides normalized identically; code spans removed before the split. A test asserts a real second sentence on the same line is still found, so the split fix cannot become a way to skip content. |
| **F-09** · P1 · correctness · `bin/dossier-claim-scan.sh`<br>The contract **mandates** that a required qualification appear beside the claim it qualifies. Qualification rows carry no `approved` cell, so the scan matched only claim rows and reported every mandated qualification as a violation. Two rules the plugin ships, contradicting each other, with no way for a drafter to satisfy both. | **Applied.** The qualification column is treated as approved text, scoped to that section only. A test asserts a rejected wording sitting in a later table is still not approved, so the section scope cannot be widened into a loophole. |
| **F-04** · P1 · correctness · `bin/dossier-claim-scan.sh`<br>The scanner read YAML frontmatter as prose, reporting `title:` and `audience:` as unregistered claims in every public document. | **Applied.** A check that emits irreducible noise trains its reader to skip its output, which is how the real findings get missed. |
| **F-10** · P2 · correctness · `tests/plugin-manifest.test.sh`<br>The licence assertion was a hard-coded `MIT` constant. A constant passes happily while the file it names says something else — which is exactly how this repository came to advertise a licence it did not carry. | **Applied.** Reads the SPDX identifier from `LICENSE` and asserts the manifest matches; fails loudly when no `LICENSE` exists. |
| **F-11** · P2 · correctness · `tests/run.sh`<br>The suite leaked ~80 temp directories per run. Test files are *sourced*, so an `EXIT` trap set by one is **replaced** by the next file's, and a trailing cleanup line strands above whatever the next contributor appends — both had already happened here, the second one to me in this cycle. | **Applied.** The runner owns one directory, points `TMPDIR` at it, removes it on exit. Verified: dossier now leaves 0 entries behind. A test asserts no test file registers its own `EXIT` trap. |
| **F-01** · P1 · correctness · 7 documents<br>The ledger recorded 11 tracked decision records; `git ls-files` returns 10. One wrong number propagated verbatim into 7 documents — the exact cross-document drift the package criticizes this repository's README for, reproduced inside the package within one drafting session. | **Applied.** Caught by the pass that re-derives counts from the filesystem rather than reading the ledger. That pass is the reason to keep method diversity even when reviewer independence is unavailable. |
| **F-02** · P1 · disclosure · `06-public/**`<br>Both public documents carried 34 inline `[CL-####]` citations. The plugin's own rule is that register identifiers must never appear publicly. | **Applied.** Removed; traceability lives in the register's mapping table, where it already was. `CLAIM_SCAN_LEAKS=0`. |
| **F-05** · P3 · method · this review<br>The three verification passes ran in one context on one model. Their agreement carries almost no independent information. | **Open, and stated rather than fixed.** The plugin makes independence architectural — separate dispatches, `memory: none`, a skill firewall. None of that was used here. Asserting independence this run did not have would be the failure the standard exists to prevent. |

### On the gate verdict

**NOT-RELEASABLE**, 15 of 17 conditions passing. Round 1 failed four; two were closed by fixing the underlying facts — the repository is now Apache-2.0 licensed with all 15 declarations aligned, and every public sentence maps to an approved claim. The two that remain are both score conditions and both trace to F-05. **No edit to the package can close them.**

The unregistered-sentence count went 52 → 0 without loosening the check: 16 sentences that genuinely assert something were registered and approved, 5 near-misses were made verbatim, and the rest was editorial glue asserting nothing about the product. Nothing was moved into a list or table to dodge the scanner — that would be gaming a check whose entire purpose is review.

### Quality gates

```
dossier  1059 assertions, 0 fail   (+25 this cycle, all regression tests for the above)
flow     1022 assertions, 0 fail
package  23/23 files · 0 header findings · LEDGER_ERRORS=0 · LEAKS=0 · UNREGISTERED=0
```

One honest caveat on that flow number: it reported `fail=4` in `flow-goal-stop.test.sh` on a single run and then passed 10 consecutive runs, including 6 back-to-back. **Not reproduced.** Recorded as EV-0055 rather than dismissed, because an intermittent failure in a suite that is already advisory is worth less than its pass count suggests. Flow's harness leaks ~254 temp directories per run from the same sourced-trap defect fixed in dossier — a plausible contributor, unproven.

danielbentes:
## Resolution — cycle 3

Nine of ten findings fix-forwarded in `f57126f` and `eab458a`. One is open by design.

| Finding | Priority | Resolution |
|---|---|---|
| F-06 | P1 | Template heading and prompt aligned to the gate; the test reads the expected string out of the gate script, so renaming either side alone fails |
| F-07 | P1 | Code spans stripped before locator classification; each span validated; loop de-piped so the error count moves with the findings it prints. `derived:` added and documented |
| F-08 | P1 | Both sides of the comparison normalized identically; code spans removed before the sentence split |
| F-09 | P1 | Required qualifications treated as approved text, scoped to that section only |
| F-04 | P1 | Header block skipped; body scanning unchanged |
| F-01 | P1 | 11 → 10 tracked decision records, corrected in all 7 documents |
| F-02 | P1 | 34 inline register identifiers removed from the public documents |
| F-10 | P2 | Licence assertion reads the SPDX identifier from `LICENSE` instead of a constant |
| F-11 | P2 | Runner owns the temp directory and removes it; verified 0 entries left behind |
| F-05 | P3 | **Open.** The passes were not independent. Recording that is the correct response; claiming independence this run did not have is the failure the standard exists to prevent |

Also in this cycle, outside the finding set: the repository is now licensed under Apache-2.0 with all 15 declarations aligned, resolving CT-0001 by evidence. AQ-0011 opened in its place and is genuinely pending — GitHub computes licence detection from the default branch, so the derived field reports it only after merge; the per-ref endpoint 404s, so branch detection cannot be observed at all.

**Gate: 15 of 17 conditions pass**, up from 13. The two failing are both score conditions, both trace to F-05, and no edit to the package can close them.

Suites: dossier **1059 / 0**, flow **1022 / 0**.

danielbentes (commented):
## Self-review — cycle 4

Paired-reviewer mode (10 agents across 5 facets, plus holdout validation in both lenses). Every finding below was re-verified in the main thread against current file state before being accepted; three were proven by executing the code rather than reading it.

**26 findings, all fix-forwarded in this PR.** Suite: dossier **1194 / 0** (from 1076), flow **1022 / 0** unaffected. Verified on macOS and in an `ubuntu:latest` container — the action-ceiling fix rewrites a command string with `sed`, so GNU-vs-BSD behaviour was tested rather than assumed.

### Previous feedback status

| Cycle | Findings | Claimed | Verified |
|-------|----------|---------|----------|
| 1 | F1–F4 | applied | confirmed |
| 2 | F5–F9 | applied | confirmed |
| 3 | F-01…F-11 (9 resolved) | applied | confirmed, **except the package-shape fix — see F-16/F-17** |
| 3 | F-05 | disputed, open by design | **still open, correctly** — see below |

Cycle 2's package-shape finding was reported resolved and pinned by a regression test. It was neither.

### P1 — Critical

| Finding | Suggested fix |
|---|---|
| **F-15 · security/containment · `hooks/scripts/enforce-output-root.sh:42`**<br>The allow-case is a `case` glob and a glob `*` matches `/`, so `docs/dossier/../../.github/workflows/evil.yml` starts with the allowed prefix and lands outside the root. Proven by executing the case statement in isolation: **allowed, exit 0**. The comment three lines below asserts the opposite. `.dossier/*` had the same hole. | Refuse any `..` segment before the prefix test. ✅ Fixed, with three traversal cases now tested. |
| **F-16 · correctness · `skills/doc-package-contract/SKILL.md:3`**<br>The skill description says `7-directory`; the Iron Law at line 15 of the same file says 8. `bin/dossier-scaffold.sh:2` and `CHANGELOG.md:19` agreed with the description. This is the line loaded into context for skill selection. | ✅ Fixed in all three. |
| **F-17 · test-integrity · `tests/package-contract.test.sh:87`**<br>**The regression test written to pin F-16 emitted PASS on the file that still carried it.** Its pattern required a space (`7 directories`) and every survivor used the hyphenated form. `grep -cE` on that exact file returns 0. `bin/` and `CHANGELOG.md` were not in the scan set at all. A green assertion certifying the bug it was written to prevent, for three rounds. | ✅ Pattern broadened, scan set widened, and the count is now **derived from the filesystem** rather than restated. Confirmed by mutation. |
| **F-18 · security/fail-open · three hooks**<br>`enforce-output-root`, `enforce-allowed-actions`, and `block-unregistered-claim` all began `command -v jq >/dev/null 2>&1 \|\| exit 0`. Absent `jq`, write containment, the action ceiling, and disclosure blocking silently no-op. | ✅ All three fail closed. `stale-header-stamp` is advisory and correctly keeps the old posture — now asserted in both directions. |
| **F-19 · security/bypass · `hooks/scripts/enforce-allowed-actions.sh:47`**<br>The boundary anchor is defeated by one layer of indirection: `/usr/bin/curl`, `env curl`, `command curl`, `bash -c "npm test"`, `eval "curl …"` all pass. | ✅ Wrappers detected and rescanned with quotes neutralised. All five now blocked; `grep -r "npm test" docs/` still passes, which is the case the anchor exists for. |
| **F-20 · security/disclosure · `hooks/scripts/block-unregistered-claim.sh:43`**<br>No `internal-path` class, though `disclosure-policy-levels.md:49` names internal repository paths as prohibited in `06-public/` **and names this hook as an enforcer**. The batch scanner had the pattern; the live block did not. | ✅ Ported from the scanner, plus an assertion that the two agree. |
| **F-21 · security · `bin/dossier-validate-patch.sh`**<br>Zero mode/symlink handling: `new file mode 120000` under an allowlisted path escapes by reference. | ✅ Refused in both modes. |
| **F-22 · security · `bin/dossier-validate-patch.sh` (found by the new test, not by review)**<br>`docs/dossier/../../.env` matches the allowlist's compiled `^docs/dossier/.*$` and was **accepted**. `git apply` would reject it without `--unsafe-paths`, but this script is the documented control and the publish job's independent re-check — it should not depend on a default in a tool someone may later invoke differently. | ✅ Traversal and absolute paths refused before the allowlist is consulted. |
| **F-23 · security · `bin/dossier-policy.sh:245`**<br>`ci.agent.model` read off the cascade unvalidated and interpolated into `claude_args`. `schema.json:567` documents the bare-token constraint *because of this risk*; the only enforcement lived in a validator the workflow never invokes. `emit()` also wrote raw values to `$GITHUB_OUTPUT`, where a newline forges output pairs for the next job. | ✅ Re-checked at the point of use; all emitted values stripped of control characters. |
| **F-24 · security · `bin/cascade-resolve.sh:81`**<br>`.claude/settings.dossier.local.json` is the top-precedence layer and was trusted on existence alone. Committed, it becomes the highest-precedence layer of settings a PR can change — silently outranking `ci.writeAllowlist`, `disclosure.policy`, and `engagement.allowedActions`. The `.gitignore` entry is a convention; `git add -f` ignores it. | ✅ Admitted only while untracked, with the skip announced. |
| **F-25 · error-handling · `templates/ci/dossier-docs-refresh.yml`**<br>Five unchecked `git` calls in a step with no `set -e`: `fetch`, `push --delete`, three `checkout -B`, and `commit` before `committed=true`. A failure left the wrong branch checked out and continued, or reported a commit that never happened. | ✅ All five checked. |
| **F-26 · correctness · `plugins/dossier/README.md:34`**<br>"Sixteen conditions … Nine are mechanical." The gate has **seventeen**, ten mechanical / seven judgment — tallied from the table the gate script implements. | ✅ Fixed, and both numbers are now derived by a test from the condition table. |
| **F-27 · correctness · `README.md:491`**<br>"four of seventeen conditions fail" — the package has said **two** since round 3, and `documentation-index.md:26` says so explicitly. | ✅ Fixed. |
| **F-28 · test-coverage · `bin/dossier-validate-patch.sh`**<br>The write-allowlist control, described in its own header as the prompt-injection backstop, had **no functional test at all**. | ✅ 14 assertions, almost all rejection cases. F-22 was found by writing them. |
| **F-29 · test-coverage · `tests/agent-independence.test.sh`**<br>The README claims a regression test enforces all five independence mechanisms. Four were asserted; per-pass model configuration was not. | ✅ Now asserted, including cascade round-trip. The README claim is now true. |
| **F-30 · test-coverage · `bin/dossier-package-check.sh`**<br>Fifteen documented finding codes, none tested — including `LEAKED-FIELD`, the check that stops an internal-only header key reaching a public document. | ✅ Fixtures for six codes plus every field on the internal-only list. |

### P2

| Finding | Suggested fix |
|---|---|
| **F-31 · error-handling · `enforce-allowed-actions.sh:19`** — `[ -x "$RESOLVER" ] \|\| exit 0` failed open where the sibling hook fails safe. | ✅ Restrictive defaults, keeps enforcing. |
| **F-32 · cleanup · `bin/dossier-evidence.sh`** — four of five temp files outside the trap; a signal or `set -u` abort leaked them. | ✅ All registered with the one trap. |
| **F-33 · hygiene · 11 sites across `bin/`** — predictable `/tmp/<name>.$$` fallback when `mktemp` fails, **including the patch validator**, while three scripts hard-failed. | ✅ All hard-fail. |
| **F-34 · correctness · `release-gate-conditions.md:5,26`, `commands/gate.md:19`, `tests/bin-scripts.test.sh:5`** — "Nine mechanical" (→ ten), "the other fifteen" (→ sixteen). | ✅ Fixed and now test-derived. |
| **F-35 · correctness · `finding-schema.md:301`, `release-gate-conditions.md:285`** — cite `AS-` and `OQ-` registers. Neither exists anywhere in the plugin; the registers are `CL-`, `CT-`, `AQ-`, `TM-`. | ✅ Fixed. |
| **F-36 · hygiene · `.flow_diff_tmp.txt`, `.handoff-pr103.md`** — 38 KB of session detritus from unrelated PRs #116 and #103, tracked, referenced by nothing, and not covered by this PR's `.gitignore` additions. | ✅ Untracked; ignore patterns added. |
| **F-37 · versioning · `plugins/ai-first-org-design-kit`** — `agent-builder/SKILL.md` got a content fix (fabricated SDK constructor, stale model ID) with no version bump, so cached installs at 1.5.0 never receive it. | ✅ Bumped to 1.5.1 across plugin.json, marketplace.json, and both READMEs. |
| **F-38 · docs-accuracy · PR body** — claimed 1001 assertions and 12 test files. | ✅ 1194 / 15, and the Claude attribution footer removed to match the repo's commit convention. |
| **F-39 · test-coverage · `tests/hooks.test.sh`** — the action ceiling's `deny()` path was never exercised; only the inert state was. | ✅ Deny path, wrapper cases, and fail-closed all tested. |
| **F-40 · test-coverage · 5 bin scripts** — hygiene checks only, including `dossier-blast-radius.sh`, whose header calls its mapping "the single biggest lever on the cost and reviewability of a refresh". | ✅ Behavioural coverage added. |
| **F-41 · lifecycle · `.scope.json`** — never removed, and the hooks key off its existence, so enforcement stays tripped for every later session. Not gitignored. | ✅ Ignore pattern added. |
| **F-42 · correctness · `tests/run.sh:112`** — a bare `2>&1` before the closing paren of a command substitution is a null command's redirection, so test stderr was never captured into `$OUTPUT`. Proven with a minimal case. | ✅ Fixed. |
| **F-43 · test-coverage · `references-integrity.test.sh:126`** — `CMD_COUNT` computed and discarded, so the README's command surface was unchecked while skills and agents were. | ✅ Asserted; heading now carries the count. |

### P3

| Finding | Suggested fix |
|---|---|
| **F-44 · `bin/dossier-pr-body.sh:76`** — escapes `\|` but not `[`, `]`, or backtick, so a hostile PR title can rewrite a table cell into a link. | ✅ Fixed and tested. |
| **F-45 · `bin/dossier-package-check.sh:138`** — no cleanup trap, unlike its three siblings. | ✅ Added. |
| **F-46 · `bin/dossier-blast-radius.sh:244`** — a failed `jq` write dropped the event silently, so the report read as "nothing matched". | ✅ Now fatal and reported. |
| **F-47 · `references/release-gate-conditions.md`** — G17's section sat after `## Related references`, out of sequence. | ✅ Moved into order. |
| **F-48 · `tests/hooks.test.sh:126`** — the stale-header test asserted only exit 0, which is that script's only outcome by design; the warning logic could be deleted and it would still pass. | ✅ Message content asserted, plus a same-day silence case. |
| **F-49 · `commands/*.md`** — `argument-hint` double-quoted in all 9; flow and decipon leave it bare. | ✅ Unquoted. |
| **F-50 · `plugins/dossier/README.md:1`** — bare lowercase H1 where siblings use a title-cased name plus descriptor. | ✅ Retitled. |

### Closed by tooling the earlier cycles could not run

`shellcheck` was unavailable in the review environment (reported as an open gap in cycle 4's test pass). Run in a container at 0.11.0 over all 33 scripts at warning severity: **one dead variable in `bin/`, three in `tests/`, and the `run.sh` redirection bug above.** All fixed; the suite is now clean at that severity and **shellcheck is a pinned CI step**, verified end to end in an `ubuntu` container rather than assumed to work.

### Assessed and deliberately not changed

- **F-05 (verification-pass independence) stays open and disputed.** It is the one finding no edit can close — closing it requires a genuinely independent re-run, and recording that honestly is the correct behaviour. Round 4 reinforces it: five containment defects were found by a reviewer reading the plugin, not by three passes over the package documenting it. The verification report now says plainly that round 4 was a fix cycle and **not** a verification pass, so a bumped round number does not imply fresh assurance.
- **Heading counts (`## Skills (9)`)** — flagged as staleness-prone. They are not: `references-integrity.test.sh` pins them to the number on disk. A mechanically-checked count is the mechanism this plugin advocates, so they stayed, and Commands gained one.
- **CHANGELOG heading format** — dossier matches `agent-capability-standard`; flow is the outlier. Not this PR's defect.
- **Branch name lacks an issue number.** Renaming mid-PR breaks the PR for no gain; noted rather than churned.

### One finding was a mislabel rather than a stale figure

`CHK-16` runs `git rev-list --count HEAD`, which counts the assessed branch, and five documents printed its result as the commit count of `main`. `main` has 189. The number was correct for what was measured and the noun described something else — nothing about it looked wrong, which is the harder version of the defect this format exists to surface.

### Verification

- dossier **1194 / 0**, flow **1022 / 0** — both on macOS and in `ubuntu:latest`
- shellcheck 0.11.0: **0 findings** at warning severity across 33 scripts
- Package self-check `CLEAN` (23/23 files, 6/6 diagrams), ledger 56 rows / 0 errors, claim scan 0 leaks / 0 unregistered sentences
- Mutation-tested: reverting each of the traversal guard, the jq fail-closed, the `internal-path` class, the hyphenated directory count, the condition count, and the mechanical split each produces a failing assertion
- The package is re-pinned to `4c1b62f` with all five commit-derived figures re-derived from it

danielbentes:
## Resolution — cycle 4

All 36 findings fix-forwarded across `d8c10fa`, `6455f54`, `031adf1`, `4c1b62f`, and `cda8620`. Nothing deferred, nothing escalated.

| Group | Findings | Resolution |
|---|---|---|
| Containment | F-15, F-18, F-19, F-20, F-21, F-22, F-23, F-24 | Traversal refused before the prefix test; three hooks fail closed; wrapper indirection defeated; `internal-path` ported from the scanner; symlink mode and traversal refused in both validator modes; `ci.agent.model` re-checked at the point of use and `$GITHUB_OUTPUT` values stripped of control characters; the project-local cascade layer admitted only while untracked |
| The test that hid a defect | F-17 | Pattern broadened to the hyphenated form, scan set widened to every file stating the count, and the count plus the gate's mechanical/judgment split are now **derived from the artifacts** rather than restated. Four mutations confirm each fires |
| Self-contradiction | F-16, F-26, F-27, F-34, F-35, F-47 | Directory count, condition count, mechanical split, failing-condition count, register prefixes, and G17's position all corrected |
| Error handling | F-25, F-31, F-32, F-42, F-45, F-46 | Five unchecked `git` calls; resolver-missing now fails safe; all temp files under one trap; `run.sh` actually captures stderr; missing trap added; dropped event now fatal |
| Coverage | F-28, F-29, F-30, F-39, F-40, F-43, F-48 | Three new test files, hook deny/traversal/fail-closed cases, the fifth independence mechanism, and the discarded command count |
| Hygiene and convention | F-33, F-36, F-37, F-38, F-41, F-44, F-49, F-50 | Predictable temp paths removed, session detritus untracked, cross-plugin version bumped, PR body corrected, `.scope.json` ignored, markdown escaping widened, frontmatter and README style aligned |

**F-05 remains disputed and open, deliberately.** No edit closes it — only a genuinely independent re-run does, and round 4 is evidence for it rather than against: five containment defects were found by a reviewer reading the plugin, not by three passes over the package that documents it. The verification report now states that round 4 was a fix cycle and not a verification pass, so the bumped round number cannot be read as fresh assurance.

**Verification.** dossier 1194 / 0 and flow 1022 / 0, on macOS and in `ubuntu:latest`. shellcheck 0.11.0 reports 0 findings at warning severity across 33 scripts and is now a pinned CI step, verified end to end in a container. Package check CLEAN, ledger 56 rows / 0 errors, claim scan 0 leaks. The package is re-pinned to `4c1b62f` with every commit-derived figure re-derived from it.

Issue #143, opened 2026-07-28
dossier: enforce-allowed-actions.sh's action-ceiling backstop can be bypassed via find -exec / xargs -I{}


danielbentes:
## Context

`plugins/dossier/hooks/scripts/enforce-allowed-actions.sh` enforces the `dossier.engagement.allowedActions` ceiling on Bash commands via two anchors: `BOUND` (a strict command-boundary anchor) and `WRAPPER` (a token list of indirection commands — `bash -c`, `env`, `timeout`, `sudo`, etc. — that, when present, switches the boundary check to "every whitespace run is a boundary" so an argument can't hide a denied keyword behind the wrapper).

Discovered as a byproduct of the code-review pass on PR for issue #137 (isolated scanner execution), which added two new deny-blocks reusing this same shared mechanism.

## Current State

Confirmed by direct reproduction: `find . -exec curl https://evil.invalid {} \;` is **not blocked** even when `networkAccess` is `false`, and the equivalent `find . -exec osv-scanner {} \;` / `xargs -I{} pyscn {}` shapes are not blocked for the two flags added in issue #137 either. `-exec` and `-I{}` are glued to a flag (`-exec`, not a standalone `exec` token; `-I{}` is one token, not `xargs` followed by a bare command word), so neither the `WRAPPER` regex nor the strict `BOUND` anchor ever fires for the denied command sitting inside them.

This is a gap in shared infrastructure that predates issue #137 (`RUN_TESTS`/`RUN_BUILD`/`NETWORK` are equally affected) — issue #137's two new blocks (`RUN_SECURITY_SCAN`/`RUN_CODE_QUALITY_SCAN`) simply inherited it unchanged.

Real-world exposure is currently low: this hook is a local/interactive backstop only. The CI `refresh` job's own `--allowedTools` allowlist (`templates/ci/dossier-docs-refresh.yml`) doesn't include `find`, `xargs`, or a bare shell at all, so the automated pipeline has no reach here regardless of this gap — only a local dossier run (a human or an agent already inside a permissive Bash context) is exposed, and even there it requires deliberately constructing a bypass rather than being reachable by accident.

## Objective

[...]

Issue #135, opened 2026-07-28
dossier: staleness should actively trigger re-verification; post-merge refresh suggestion must not depend on another plugin


danielbentes:
## Context

The dossier plugin is an evidence-first documentation-generation system for Claude Code. It is meant to keep a generated documentation package current as the underlying project changes — including staying current purely with the passage of time, since real-world facts can drift even when no related code changes.

This is Part 1 of a four-part sequential initiative closing verified gaps between what dossier already produces and a genuinely continuously-current documentation package. Each part ships and is proven independently before the next starts; parts 2-4 (vulnerability-evidence ingestion plus a new release-gate condition, isolated vulnerability/code-quality scanner execution, and a telemetry-only rolling-branch-rotation observability pass) will be filed as separate issues.

## Current State

Two verified, real gaps in dossier's post-merge continuous-update automation:

1. Staleness detection exists only as a reporting signal today. A document can sit well past its configured staleness threshold with no related code change nearby, and nothing about that ever triggers a refresh — a maintainer has to notice it manually.
2. There is a "suggest a documentation refresh after a local merge" setting, but nothing in the product actually implements it today. Wiring it up the way it was originally scoped would require dossier to depend on a specific other plugin being installed (and that plugin to depend on dossier back) — which conflicts with this marketplace's requirement that every plugin work correctly entirely on its own, with no other plugin present, in either direction.

## Objective

- A document that has gone stale purely from time passing (no matching code change) should, on a bounded, predictable cadence, receive a real re-verification pass — not just a report a human has to notice. Verification must never silently turn into an unprompted rewrite: if the underlying facts still check out, only the "last verified" signal should move.
- A local merge to the default branch should be able to prompt (or, if configured, automatically start) a documentation refresh, regardless of what tool or workflow performed that merge — and this must work with only the dossier plugin installed, with no other plugin required in either direction.
- Any duplicate or inconsistent staleness logic living in more than one place today should be consolidated into a single source of truth, so future changes can't cause different consumers (status reporting, refresh planning, etc.) to quietly disagree with each other.

## Acceptance Criteria

- [ ] A document that has crossed its configured staleness threshold, with no other trigger present, results in an actual re-verification pass on a scheduled cadence — demonstrated by a reproducible test scenario, not just documentation.
- [ ] A re-verification triggered purely by staleness never rewrites a document whose claims are still accurate — only its "last verified" signal advances.
[...]

PR #144, opened 2026-07-29
feat(dossier): execute isolated dependency-vulnerability and code-quality scans

### LINKED ISSUE #137: dossier: execute dependency-vulnerability and Python code-quality scans, isolated within the automated refresh pipeline's security boundaries
## Context

Part 2 of this initiative (tracked separately) gives dossier a way to cite vulnerability evidence a project has already produced. This issue, Part 3, is the natural follow-on: letting dossier generate that evidence itself, by actually running a dependency-vulnerability scan and a Python code-quality scan, for projects that don't already have that output available.

An earlier design for this capability assumed it could run from inside dossier's existing automated documentation-refresh workflow, using the same execution environment that workflow already uses for its language-model-driven steps. That assumption was checked against the workflow's actual, deliberately restrictive execution environment and does not hold: that environment is intentionally locked down to a narrow, pre-approved set of operations, specifically so that an automated documentation update can never execute arbitrary external tools or reach the network unexpectedly. Any new scanning capability has to work within that constraint, not around it.

## Current State

Dossier cannot currently execute a vulnerability or code-quality scan itself under any circumstance — this is a genuine gap, not a configuration option waiting to be enabled. Running one requires launching a new external tool against the target project's source, in some configurations reaching the network, inside an automated pipeline that currently has no execution path designed for that.

## Objective

- A project owner should be able to opt in to dossier actually running a dependency-vulnerability scan and, for Python projects, a code-quality scan, rather than relying only on scan output the project already has.
- This execution must happen in a way that respects the automated pipeline's existing security boundaries: it should not gain broader tool or network access than the specific scan requires, and it should not run with credentials or write access it doesn't need.
- The two kinds of scan (dependency vulnerabilities vs. Python code quality) must be clearly separate, opt-in capabilities — enabling one must never silently enable the other, since they carry different scope and different risk.
- When a scan cannot genuinely run — the capability isn't enabled, the tool isn't available, or (for offline vulnerability checks) the local reference data is missing or too old to trust — the result must say so honestly. It must never be indistinguishable from "ran, and found nothing."
- This capability should have a defined, bounded cost (time, data downloaded, compute) rather than an open-ended one.

## Acceptance Criteria

- [ ] With the capability enabled, dossier produces real vulnerability-scan results for a project with known vulnerable dependencies, correctly cited as evidence.
- [ ] With the capability enabled on a Python project, dossier produces real code-quality scan results.
- [ ] Enabling one of the two scan types does not enable the other.
- [ ] A scan that cannot genuinely run for any reason (capability disabled, tool unavailable, stale or missing offline reference data) is clearly and distinctly reported as not having run — never reported as a clean result.
- [ ] The scanning capability operates within the automated pipeline's existing security boundaries (no broader network, credential, or write access than the specific scan needs).
- [ ] At least one real end-to-end run against the actual scanning tools (not a substitute/stand-in) confirms the capability works in practice, not just in isolated tests.
- [ ] The full dossier test suite passes.

danielbentes:
## Summary

Closes #137. Adds isolated, deterministic execution of dependency-vulnerability (`osv-scanner`) and Python code-quality (`pyscn`) scans, gated independently by `dossier.engagement.allowedActions.runSecurityScan` / `runCodeQualityScan`, both defaulting `false`. Scanners never run inside the LLM agent's own Bash tool — they run as a standalone pre-agent CI step (`plugins/dossier/templates/ci/dossier-docs-refresh.yml`'s new `scan` job) with read-only permissions and no LLM credential, and locally via `dossier-scan-security.sh` / `dossier-scan-quality.sh`.

## FlowGoal Status

- **Goal**: `issue-137` (`.flow/goals/issue-137.goal.yaml`)
- **Lifecycle**: `achieved` (confidence 0.88)
- **Evidence**: per-criterion FlowEvidence sidecars under `.flow/runs/2026-07-28T172831Z-issue-137/evidence/`

All 7 acceptance criteria pass:

| # | Criterion | Verification |
|---|---|---|
| AC1 | Real vuln-scan results for known-vulnerable deps, cited as evidence | `vuln-scan-execution.test.sh` (real osv-scanner 2.4.0 run, 12 real Critical/High CVEs) |
| AC2 | Real code-quality results on a Python project | `quality-scan-execution.test.sh` (real pyscn 1.10.2 run) |
| AC3 | Enabling one scan type never enables the other | tripwire assertions in both scan-execution test files |
| AC4 | Never reported clean when it can't genuinely run | disabled/unavailable/timeout/offline-missing/offline-stale/offline-empty-dir cases |
| AC5 | Operates within the pipeline's existing security boundaries | `hooks.test.sh` deny-blocks + `workflow-template.test.sh` CI job permission/credential/pinning assertions |
| AC6 | At least one real end-to-end run against actual tools | satisfied by AC1/AC2's real local runs against installed osv-scanner/pyscn |
| AC7 | Full dossier test suite passes | `bash plugins/dossier/tests/run.sh` |

## Review cycle

Full `/flow:pr` Phase 3 fan-out (code-reviewer, convention-checker, test-runner, security-reviewer, error-handler-inspector, holdout-validation, integration-verifier) ran across two rounds:

**Round 1** — code-reviewer found 5 findings (P1:1, P2:2, P3:2), all fixed:
- F1 (P1): a non-object element in osv-scanner's `vulnerabilities[]` silently discarded an entire package's findings — fixed with per-element filtering + a separate `unparseable()` marker.
- F2 (P2): the CI refresh job was implicitly success-gated on the scan job — fixed with `always()` + `continue-on-error`.
- F3 (P2): a pre-existing `find -exec`/`xargs -I{}` bypass in `enforce-allowed-actions.sh`, inherited unchanged — documented, not patched (matcher-narrowing lesson), filed as #143. Low exposure: the CI refresh job's tool allowlist contains neither `find` nor `xargs`.
- F4/F5 (P3): silent failure paths in both wrappers now warn/create their fallback directories.

Deeper testing during this round also surfaced and closed an AC4 gap: osv-scanner has no staleness check on its own cached database (verified live — a database aged to 2020 loaded with zero warning). `dossier-scan-security.sh` now checks cache age independently before invoking the tool.

**Round 2** — holdout validation (cross-referencing self-review claims against actual file state) found 5 more findings (P1:1, P2:2, P3:2), all fixed:
- P1: the offline-staleness guard keyed its age check on the *newest* file across the whole cache tree, not the ecosystem subdirectory the target actually needs — a fresh, unrelated ecosystem could mask an arbitrarily stale one. First fix keyed on the *oldest* file instead (fail-closed) — see Round 3, this didn't hold up.
- P2: the CI cleanup step deleted `osv-scan-raw.json` before upload, but the security envelope (unlike quality's) embeds no findings of its own — only `artifact_path` naming that now-deleted file. Fixed to retain the security raw file; only the genuinely redundant `pyscn-scan-raw.json` is dropped. This is a **partial, disclosed revert** of this same PR's earlier SEC-5 fix, for the security half only — SEC-5's premise ("nothing in this repo reads these raw files directly") was true for `pyscn-scan-raw.json` but false for `osv-scan-raw.json`, which is the only input `dossier-vuln-evidence.sh --scan` or a downstream `Read` can get citable vulnerability content from. The envelope's `note` field, uploaded in the same bundle, is the disclosure framing SEC-5 was protecting; pyscn's raw file (self-sufficient envelope, no downstream consumer) is still dropped.
- P2: the test covering that cleanup step matched a comment, not the real `rm` step — fixed to anchor on the literal step line.
- P3 x2: a stale diagnostic message naming an old fixture's packages; a silent fallthrough on an empty-but-existing cache directory (resolved as a side effect of the P1 fix).

**Round 3** — an advisor consultation before dispatching `/flow:review` caught that Round 2's P1 fix (whole-tree "oldest file") traded one wrong failure mode for another: a machine with a genuinely long-lived, multi-ecosystem cache would have its `--offline` capability permanently refuse a fresh, actively-used ecosystem forever, the moment *any* unrelated sibling ecosystem's cache aged past the threshold. Confirmed as real (not theoretical) by writing the inverse case — fresh PyPI db, stale unrelated npm sibling, PyPI target — and observing it RED against the whole-tree code. Redesigned to check staleness **per ecosystem, after invocation**, using osv-scanner's own JSON output as ground truth for which ecosystem(s) it actually consulted (`results[].packages[].package.ecosystem`, verified live for both PyPI and npm targets) — no manifest-to-ecosystem guessing needed, since osv-scanner already resolves ecosystem from the lockfile before touching the vulnerability cache. Six regression cases added/updated, RED→GREEN verified where a prior baseline existed.

All fixes TDD RED→GREEN verified individually. Full targeted suite (23 files, excluding two pre-existing files unrelated to this diff with a known local-git-fixture-leak bug from #135) re-run clean after every round: **1673/1673**, shellcheck -S warning clean, `bash -n` clean.

Integration verification (independent re-run + a manual spot-check invocation outside the test harness) returned zero findings.

## Known, disclosed limitations (non-goals, not gaps)

- The CI scan job's isolation (permissions, credential absence, pinned/checksummed installs) is validated structurally (YAML-parse + assertions), not via an actual live GitHub Actions run.
- Issue #143 (find/xargs hook-bypass) is a pre-existing, documented limitation this PR inherits but does not widen.
- The offline per-ecosystem staleness check trusts osv-scanner's own reported `ecosystem` field as ground truth for what was consulted; it does not independently re-derive ecosystem from the target's lockfiles.

## Test plan

- [x] `bash plugins/dossier/tests/run.sh` — full targeted suite (23/25 files) green, 1673/1673
- [x] `shellcheck -S warning` clean on all modified/new bash scripts
- [x] `bash -n` clean on all modified/new bash scripts
- [x] Real end-to-end runs against actually-installed osv-scanner 2.4.0 and pyscn 1.10.2
- [x] Holdout validation and integration verification both closed with zero remaining findings


danielbentes (commented):
## Self-Review Summary

Performed self-review with fix-forward, via `/flow:review`'s Path A paired-reviewer protocol (10 agent invocations across 5 facets + 2 holdout-validation lenses).

### Verdict Summary

Verdict: N/A (independent verdict-judge not part of this review pass — that ran separately as part of the FlowGoal evaluation, verdict `achieved`, confidence 0.88).

### Findings Found & Fixed

| Finding | Fix Applied |
|---------|-------------|
| **F1 · P1 · `dossier-scan-security.sh:339`**<br>Per-ecosystem offline staleness check never fires on a clean scan — real osv-scanner only lists packages with findings in `results[].packages[]`, so a clean scan produces `results:[]` and the check silently never runs. Found independently by 4 reviewers with live reproduction against real osv-scanner. | Added `--all-packages` to the offline scan invocation, forcing every resolved package to be listed regardless of vulnerability status. Verified this changes nothing else observable. Closed with an argv-capture tripwire and a real end-to-end test against a genuinely clean dependency. |
| **F2 · P1 · `dossier-scan-security.sh:146`, `dossier-scan-quality.sh:143`**<br>`emit()`'s `--out` write-failure branch exited before printing the already-computed JSON envelope to stdout, breaking the documented exit-code contract. Reproduced by 2 reviewers. | Print `$RESULT` unconditionally before attempting the `--out` write; a failed write is now a stderr warning, not a fatal exit. |
| **F3 · P2 · `dossier-scan-security.sh:113,230`, `dossier-scan-quality.sh:90`**<br>Malformed `DOSSIER_SCAN_TIMEOUT_SECONDS`/`DOSSIER_SCAN_OFFLINE_MAX_AGE_DAYS` silently disabled the timeout guard or crashed with `set -u`. Reproduced by 2 reviewers. | Both env vars now validated with a fallback to their documented defaults on non-numeric input. |
| **F4 · P2 · `dossier-docs-refresh.yml` scan job**<br>The 4 scan-job steps had no `always()`/`continue-on-error` between them, so a hard failure in one silently skipped the rest. Found independently by 3 reviewers. | Added `if: always()` to all 4 steps, matching the `refresh` job's own existing pattern. |
| **F5 · P2 · `enforce-allowed-actions.sh:86`**<br>`python3 -m pyscn` bypassed the `runCodeQualityScan` deny check (pyscn is pip-installed, an ordinary invocation shape). Reproduced empirically. | Added `python`/`python3` to the `WRAPPER` indirection-token list. |
| **F6 · P2 · `dossier-vuln-evidence.sh:295`**<br>Alias-summary lookup took the first array-order matching id's summary rather than the first non-empty one — 4 of 12 real findings in this issue's own e2e fixture shipped a blank summary. | Filter out no-summary candidates before selecting first, instead of coercing null to `""`. |
| **F7 · P2 · `skills/evidence-ledger/SKILL.md:83`**<br>Stale text ("Dossier never executes a scanner") never referenced the new CI-produced `.dossier/scan/` bundle — the doc that actually drives the refresh agent's behavior. | Updated to reference both the CI scan bundle and pre-existing project artifacts. |
| **F8 · P3 · `dossier-scan-security.sh:104`**<br>`--help` sed range had gone out of sync with the header (truncating before the exit-code contract). | Widened the range to match the current header length. |
| **F9 · P3 · `dossier-scan-security.sh:201`, `dossier-scan-quality.sh:191`**<br>`mktemp -d` fallback path (only reached when `mktemp` itself fails) left a predictable, unrestricted `/tmp` directory. Found by 2 reviewers. | `chmod 700` the fallback directory after creation. |
| **F10 · P3 · `dossier-scan-security.sh:315`, `dossier-scan-quality.sh:213`**<br>Neither script trapped `INT`/`TERM`, so a locally-interrupted wrapper never signaled its backgrounded tool child. | Both scripts now trap and kill the child, clearing the trap on normal exit. |
| **F11 · P1 · git history (11 of 24 commits)**<br>Commit authorship corrupted to `Test <test@example.com>` — traced to a local `.git/config` override in this actual repository (likely the same unexplained isolation failure behind an earlier-session git-corruption incident). Found independently by 2 reviewers. | User unset the local git config override; 11 commits rebased with `git commit --amend --author=...` to restore correct authorship. Verified via `git log --format='%an <%ae>'` across the full range before force-pushing. |
| **F12 · P2 · git history (`f040a74`)**<br>A literal heredoc `EOF` delimiter leaked into a commit message body. Found independently by 2 reviewers. | Fixed via a scripted, non-interactive `git rebase -i` rewording the one affected commit. |
| **F13 · P2 · git history (`79209de`)**<br>Commit subject broke the `type(scope):` convention every other commit on the branch follows. Found independently by 2 reviewers. | Fixed in the same scripted rebase as F12. |

### Verification
- [x] Quality commands pass after fixes — full targeted suite (23 files, excluding 2 pre-existing files unrelated to this diff with a known issue #135 git-fixture-leak bug): **1703/1703**, `shellcheck -S warning` clean, `bash -n` clean
- [x] No new issues introduced — re-ran the full suite after every commit in this round
- [x] All P1/P2/P3 findings in touched files fixed in-PR (none escalated)

danielbentes:

### Review Cycle Summary

| Metric | Count |
|--------|-------|
| Received | 13 |
| Fixed | 13 |
| Discussed | 0 |
| Escalated | 0 |

### Changes Made

**1. F1 — offline staleness check never fires on a clean scan**
- Added `--all-packages` to the offline scan invocation so osv-scanner lists every resolved package regardless of vulnerability status
- Commit: `5e33763`

**2. F2 — `--out` write failure suppressed the JSON envelope**
- `emit()` now prints the result before attempting the optional `--out` write in both scripts
- Commit: `5e33763`

**3. F3 — malformed timeout/max-age env vars broke the exit-code contract**
- Both env vars validated with fallback to documented defaults
- Commit: `5e33763`

**4. F8, F9, F10 — `--help` truncation, mktemp fallback perms, interrupted-wrapper trap**
- `--help` range widened; fallback scratch dirs `chmod 700`; both scripts trap `INT`/`TERM`
- Commit: `5e33763`

**5. F4 — CI scan-job steps not fault-isolated**
- `if: always()` added to all 4 scan-job steps
- Commit: `09ec6be`

**6. F5 — `python3 -m pyscn` bypassed the deny backstop**
- `python`/`python3` added to the `WRAPPER` indirection-token list
- Commit: `f597248`

**7. F6 — vuln-evidence.sh picked the wrong alias summary**
- Filter out no-summary candidates before selecting first
- Commit: `5d0aa2d`

**8. F7 — evidence-ledger SKILL.md not updated for the new scan capability**
- Now references the CI-produced scan bundle
- Commit: `3bc40e2`

**9. F11, F12, F13 — git history integrity (commit authorship, stray heredoc EOF, missing commit scope)**
- Local `.git/config` identity override unset (user action); 11 commits rebased to correct authorship; 2 commits rebased to fix the stray `EOF` and missing scope
- Verified via `git log --format='%an <%ae>'` across the full range before force-pushing with `--force-with-lease`

### Thread Status

| Thread | Status |
|--------|--------|
| **F1–F13, self-review findings** | Resolved |

### Verification

- [x] All quality checks pass — full targeted suite 1703/1703, `shellcheck -S warning` clean, `bash -n` clean
- [x] All P1/P2/P3 findings in touched files addressed in-PR

PR #145, opened 2026-08-02
feat(dossier): observe rolling-branch rotation status, telemetry only

### LINKED ISSUE #138: dossier: make rolling-branch rotation observable (telemetry only) before any automated branch action is considered
## Context

Dossier's automated documentation-refresh pipeline accumulates its generated changes on a single long-lived branch and pull request over time, rather than opening a fresh one for every update. A configuration option already exists implying that branch should periodically "rotate" to a fresh one, so the accumulated pull request never grows unreviewably large — but nothing currently implements that rotation.

An earlier design for implementing it involved the pipeline autonomously closing the current pull request and deleting its branch on a schedule, before opening a replacement. That design was checked against the pipeline's actual safety invariants and found to have a real failure mode: the existing logic that guards against exactly this kind of destructive action is explicitly built around the assumption that nothing like it happens while a pull request is open, and the safeguard meant to prevent repeated/runaway action can silently disable itself if a lookup it depends on fails. In the worst case, a bug could cause the pipeline to close and recreate its own pull request repeatedly, discarding review history and comments each time, with no human in the loop to notice before it happens again.

Given that, this issue is deliberately scoped to only the safe, non-destructive first step: observing when rotation *would* happen, without ever actually taking a destructive action. Actually rotating branches — and definitely deleting one — is out of scope here and would need its own dedicated proposal, backed by real observation data from this issue's work, before it's ever considered.

## Current State

There is no visibility today into whether or how often the documentation pipeline's accumulating branch would benefit from rotation. The relevant configuration option exists but has no effect.

## Objective

Make the rotation *decision* observable — without ever acting on it. Every pipeline run should be able to determine and record whether rotation would be warranted at that point (based on time elapsed and/or how large the accumulated, unreviewed changes have become), and why, purely as a visible log entry a maintainer can review. No branch is ever closed, deleted, or recreated as part of this work.

## Acceptance Criteria

- [ ] Every pipeline run records a clear, human-readable determination of whether branch rotation would currently be warranted, and the specific reason.
- [ ] No pipeline run, under any circumstance, closes a pull request, deletes a branch, or creates a replacement branch as a result of this capability — it is purely observational.
- [ ] The determination logic is exercised by tests covering both "would rotate" and "would not rotate" scenarios.
- [ ] After a period of real observation on this repository's own pipeline runs, there is enough recorded data to make an informed decision about whether an actual (non-destructive-by-design) rotation mechanism is worth proposing as separate, future work.
- [ ] The full dossier test suite passes.

danielbentes:
## Summary

Closes #138. Adds `dossier-rotation-check.sh`, a new, purely-observational telemetry script that determines whether dossier's rolling documentation branch would currently warrant "rotation" — without ever taking any action. An earlier design (autonomous close/delete/recreate on a schedule) was rejected as unsafe during issue design because the existing destructive-action guard assumes nothing like it happens while a PR is open, and its own circuit breaker can silently disable itself if a lookup fails. This issue is scoped to only the safe first step: observe and log, never act.

## What changed

- **New**: `plugins/dossier/bin/dossier-rotation-check.sh` — resolves the rolling branch's age (open PR's `createdAt`, falling back to the oldest `Dossier-Generated: true` commit) and cumulative diff size against the base branch, computes a tri-state `would_rotate` (`true`/`false`/`unknown` — never silently coerced to a confident "false" when the truth is unmeasured), and emits both a `key=value` output and a human-readable job-summary table.
- **CI wiring**: a new, unconditional step inside the `policy` job of `dossier-docs-refresh.yml`, running on every trigger (including declined runs), entirely within the job's existing `contents: read`/`pull-requests: read` scope — no new permission.
- **Config**: new `dossier.ci.thresholds.rotationMaxAccumulatedLines` field (default 5000) in `schema.json`/`settings.json`/`templates/config.example.json`.
- **Tests**: new `rotation-check.test.sh` (57 assertions against real local git repos with a real bare-repo remote), plus scoped safety assertions in `workflow-template.test.sh` and a `bin-scripts.test.sh` registration.

## Acceptance criteria

AC4 as originally written ("after a period of real observation... enough data to decide") described a future outcome no command could verify today. Reworded and user-confirmed before implementation: metrics (`age_days`/`accumulated_files`/`accumulated_lines`) are always computed and emitted in a stable format across every state — including when rotation is disabled — so a real observation period actually produces usable data. Full rationale in `.decisions/issue-138.md`.

| # | Criterion | Verification |
|---|---|---|
| 1 | Every run records a determination + reason | `rotation-check.test.sh` + `workflow-template.test.sh` |
| 2 | Never closes a PR / deletes / recreates a branch | scoped negative assertions against the script body + the `policy` job's own YAML block (never a whole-file scan, since `publish`'s own pre-existing destructive-guard code legitimately contains those verbs) |
| 3 | Tests cover both directions | `rotation-check.test.sh` — would-rotate via age, via size, via degraded fallback; would-not-rotate |
| 4 (reworded) | Metrics always emitted, every state | `rotation-check.test.sh` — explicit non-empty assertions under `rollingBranchRotation=none`, explicit empty (not zero) assertions on cold start |
| 5 | Full suite passes | `bash plugins/dossier/tests/run.sh` — 1802/1802 |

FlowGoal `issue-138` reached `achieved` (confidence 0.85) via two rounds with an independent goal-evaluator-judge, which caught and required closing two real evidence gaps before returning a verdict.

## Review round (5 parallel agents: code quality, security, convention, test-runner, error-handling)

| Finding | Priority | Fix |
|---|---|---|
| New CI step had no `continue-on-error`; its failure could cascade into skipping the real evidence-bundle build for a legitimate merge | P1 | Added `continue-on-error: true`, matching the file's own existing precedent for advisory steps |
| An explicitly-empty `rollingBranch` config override collides with `git ls-remote`'s "branch doesn't exist" exit code (both are 2), producing a confidently wrong result — reproduced live | P1 | Value guards on `DOCS_BRANCH`/`ROTATION_POLICY`, falling back to documented defaults |
| `gh pr list` failures (auth/rate-limit) were indistinguishable from "no PR found" | P2 | Now logged as a distinct degradation note |
| Job-summary markdown cells had no escaping — a value with a backtick/pipe/newline could break a table or inject a spoofed row | P2 | Added `sanitize_md()` at the display boundary only |
| `--help`'s line range included one line past the real header | P3 | Fixed |
| An unused captured variable (shellcheck SC2034) | P3 | Now asserted on, strengthening the test |
| `DOCS_BRANCH` has no shape guard against a leading-dash flag-injection risk | P3 | **Disclosed, not fixed** — inert in the shipped CI workflow (HTTPS remote); a real (if narrow) risk only for standalone/local use with a non-HTTPS origin. The reviewer's own suggested fix needed verification of `git ls-remote`'s `--` handling first, which wasn't done — left as a known limitation rather than guessed at. |

All P1/P2 fixes were RED→GREEN verified (reproduced the failure against pre-fix code, then confirmed the fix). Full suite re-confirmed at 1802/1802 after every round.

### Out-of-band incident (not a code finding)

Mid-review, a pre-existing, unrelated test-isolation bug in `plugins/dossier/tests/local-merge-hook.test.sh` (its git fixture setup, when it fails to isolate, runs bare `git config`/`git checkout`/`git commit` against the real repository rather than its intended `mktemp` sandbox) corrupted this session's local working directory twice: once creating stray local branches (`conflict-branch`, `feature-branch`, never pushed, harmless), and once overwriting the local `.git/config` identity to the fixture's `Test <test@example.com>`, which briefly misattributed 5 already-committed commits on this branch. Both were fully diagnosed (via reflog and exact fixture-string matching) and repaired — the stray branches via `git branch -f` to known-good refs, the misattributed commits via a non-destructive `git rebase --exec 'git commit --amend --author=...'` (nothing had been pushed at the time). All 10 commits on this branch are correctly attributed; content integrity reconfirmed via a fresh full-suite run after each repair. **This bug itself is out of scope for this PR** (a separate, pre-existing file unrelated to rotation-check telemetry). Root-caused and filed as [#149](https://github.com/synaptiai/synapti-marketplace/issues/149): two call sites capture `_dossier_safe_mktemp_dir`'s output via `$(...)` without checking the exit code, so a resolver/mktemp failure silently yields an empty path — and `cd ""` returns exit 0 in bash, turning the fixture's own isolation guard into a no-op.

## Review cycle 1 (Path A: paired-reviewer + challenge protocol, 32 findings)

A full Path A review (5 facets × skeptic/verifier + holdout ×2) found 32 consolidated findings beyond the 5-agent round above. All 32 were dispositioned: **14 fixed in-PR** (RED→GREEN verified), **4 filed as separate tracked issues** ([#146](https://github.com/synaptiai/synapti-marketplace/issues/146), [#147](https://github.com/synaptiai/synapti-marketplace/issues/147) ×2, [#148](https://github.com/synaptiai/synapti-marketplace/issues/148) — each requires a CI-credential-policy decision or touches code outside this PR's diff), **14 left as documented P3 cosmetic/coverage notes**. Full per-finding breakdown in the [review comment](https://github.com/synaptiai/synapti-marketplace/pull/145#pullrequestreview-4840049406) and [resolution comment](https://github.com/synaptiai/synapti-marketplace/pull/145#issuecomment-5160981968).

Notably this cycle: a live-verified bug where `rotationMaxAccumulatedLines=0` made the size signal trigger on every run (0 is always `>=` any measured value); a fork-PR-hijack risk in the new script's `gh pr list --head` lookup (unscoped by repo — fixed here, and the same pre-existing weakness in `dossier-policy.sh` filed as #146 since it has privileged, write-token consequences there); and a `cfg()` bug that masked genuine config-resolver crashes as ordinary empty-value fallbacks (live-verified on Ubuntu) — whose first fix attempt was itself caught as broken by its own regression test before shipping (a `die_infra()` call inside a `$(...)` subshell, whose `exit` only killed the subshell).

## Verification

- [x] Cycle-1 fixes: `bash plugins/dossier/tests/run.sh rotation-check.test.sh workflow-template.test.sh` — 177/177 (`rotation-check.test.sh` alone: 81/81, up from 65 pre-cycle)
- [x] `shellcheck -S warning`: clean on the two files this cycle touched (`dossier-rotation-check.sh`, `rotation-check.test.sh`)
- [x] `bash -n`: clean
- [x] Full suite (`bash plugins/dossier/tests/run.sh`) was **not** re-run this cycle, deliberately — `local-merge-hook.test.sh` has corrupted this session's working directory twice (see above, now filed as #149); the earlier "Full suite: 1802/1802" line above reflects the pre-cycle-1 state, before this cycle's fixes
- [x] Every code/behavior fix this cycle (not the test-coverage-only additions) was reproduced RED against the pre-fix code, then confirmed GREEN
- [x] Real end-to-end runs against this actual repository (not just fixtures) at multiple points, including after every review-round fix
- [x] Holdout validation: PASS, no conflicts
- [x] Independent verdict-judge: PASS (5/5 criteria)

### DIFF
--- a/.decisions/issue-138.md
+++ b/.decisions/issue-138.md
@@ -36,6 +36,12 @@ artifacts:
   workflow: start-issue
   run_id: 2026-07-29T115831Z-issue-138
   status: completed
+- type: review-cycle
+  captured_at: '2026-08-02T23:11:01Z'
+  cycle: 1
+  path: B
+  findings_count: 8
+  pr: 145
 ---
 # Issue #138 — Rolling-branch rotation, telemetry only
 
@@ -132,3 +138,5 @@ PASS — 3 tasks reviewed (implementation-planner agent). Task 59 (script + work
 <!-- auto-log: 2026-08-03 01:02 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/rotation-check.test.sh -->
 
 <!-- auto-log: 2026-08-03 01:06 commit "fix(dossier): guard empty config overrides, sanitize job-summary output" -->
+
+<!-- auto-log: 2026-08-03 01:10 Write /private/tmp/claude-501/-Users-danielbentes-synapti-marketplace/8c76ed85-3c0c-4a32-8924-be0cf2c7bc2d/scratchpad/pr-body-138.md -->
--- a/.decisions/issue-138.md
+++ b/.decisions/issue-138.md
@@ -184,3 +184,19 @@ PASS — 3 tasks reviewed (implementation-planner agent). Task 59 (script + work
 <!-- auto-log: 2026-08-03 01:52 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/bin/dossier-rotation-check.sh -->
 
 <!-- auto-log: 2026-08-03 01:52 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/bin/dossier-rotation-check.sh -->
+
+<!-- auto-log: 2026-08-03 01:53 commit "fix(dossier): route config-resolver infra failures through die_infra" -->
+
+<!-- auto-log: 2026-08-03 01:54 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/bin/dossier-rotation-check.sh -->
+
+<!-- auto-log: 2026-08-03 01:55 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/rotation-check.test.sh -->
+
+<!-- auto-log: 2026-08-03 01:57 commit "fix(dossier): route config-resolver infra failures through die_infra" -->
+
+<!-- auto-log: 2026-08-03 01:57 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/rotation-check.test.sh -->
+
+<!-- auto-log: 2026-08-03 01:58 commit "fix(dossier): route config-resolver infra failures through die_infra" -->
+
+<!-- auto-log: 2026-08-03 01:58 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/rotation-check.test.sh -->
+
+<!-- auto-log: 2026-08-03 01:59 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/rotation-check.test.sh -->
--- a/.decisions/issue-138.md
+++ b/.decisions/issue-138.md
@@ -1,0 +1,87 @@
+---
+issue: 138
+created: '2026-08-02T20:53:11Z'
+artifacts:
+- type: specification
+  captured_at: '2026-08-02T20:53:11Z'
+  by: specification-capture
+  elements:
+  - non-goals
+  - failure-modes
+  - interface-contracts
+- type: goal-created
+  captured_at: '2026-08-02T20:53:11Z'
+  goal_id: issue-138
+  source: github_issue:138
+- type: workflow-run
+  captured_at: '2026-08-02T20:53:12Z'
+  workflow: start-issue
+  run_id: 2026-07-29T115831Z-issue-138
+  status: active
+- type: stranger-test
+  captured_at: '2026-08-02T20:53:12Z'
+  result: PASS
+  task_count: 3
+- type: verdict
+  captured_at: '2026-08-02T20:53:12Z'
+  result: PASS
+- type: goal-evaluation
+  captured_at: '2026-08-02T21:11:30Z'
+  goal_id: issue-138
+  result: achieved
+  evidence_bundle: .flow/runs/2026-07-29T115831Z-issue-138
+  failures: none
+---
+# Issue #138 — Rolling-branch rotation, telemetry only
+
+**Title**: dossier: make rolling-branch rotation observable (telemetry only) before any automated branch action is considered
+**Labels**: enhancement, plugin
+
+## Specification
+
+_Captured by specification-capture skill on 2026-07-29. Source: mixed (non-goals extracted-from-issue; failure modes and interface contracts drafted and user-confirmed)._
+
+### Non-goals
+
+- Never close a pull request, delete a branch, or create a replacement branch, under any circumstance, as a result of this capability (issue body, AC2, verbatim).
+- Never implement the actual rotation mechanism (closing/deleting/recreating the accumulating branch) — that requires its own dedicated proposal, backed by real observation data from this issue's work (issue body, "Context").
+- Never modify, weaken, or bypass the existing destructive-action guard in the `publish` job's "Prepare the documentation branch" step (the `FOREIGN`-commit walk / `EXISTING_PR` check that gates `git push origin --delete`, `dossier-docs-refresh.yml:738-822`) — this issue's script and workflow step are read-only additions to the `policy` job and must never touch the `publish` job's write-permission code path.
+- Never grant the new step more than the `policy` job's existing `contents: read` / `pull-requests: read` permission scope.
+
+### Failure modes
+
+- **Timeouts** — none — the new script performs only local `git` operations plus a single optional `gh pr list`/`gh pr view` call under the same network conditions as the existing `policy` job's `EXISTING_PR` lookup (`dossier-policy.sh:390`); no new timeout surface is introduced beyond what that job already tolerates.
+- **Partial failures** — when the `gh` CLI is absent or its API call fails, the script degrades to a git-only age determination (oldest `Dossier-Generated: true` commit on the docs branch) rather than failing outright, and logs an explicit `note()` (following `dossier-policy.sh:451/461`'s convention) rather than the silent-empty pattern the issue implicitly reacts to (`dossier-policy.sh:390`, `dossier-docs-refresh.yml:720`). If BOTH the PR lookup and the commit-based fallback are unable to produce an age, `would_rotate` is reported as the literal string `unknown`, never coerced to `false`.
+- **Invalid input** — a non-numeric or missing `dossier.ci.thresholds.rotationMaxAccumulatedLines` config value falls back to the documented default (5000) rather than crashing the script or corrupting the `--github-output` line format, matching the malformed-env-var handling precedent established in `dossier-scan-security.sh`/`dossier-scan-quality.sh` (issue #137, PR #144).
+- **Missing context** — when the rolling documentation branch does not exist on the remote at all (`git ls-remote --exit-code --heads origin "$DOCS_BRANCH"` fails), the script reports a distinct `no-branch` cold-start state — `would_rotate=false` with an honest, specific reason ("the rolling branch does not exist yet; nothing to rotate") — never silently reporting "0 lines changed / would not rotate" as if a branch had been measured and found small.
+
+### Interface contracts
+
+- New script `plugins/dossier/bin/dossier-rotation-check.sh`. CLI contract mirrors `dossier-staleness-check.sh`/`dossier-policy.sh`: `[--github-output <file>] [--summary <file>]`. Untrusted values (`BASE_REF`, `GH_TOKEN`) bound via environment only, never argv. Exit 0 = a decision was reached (`would_rotate` may be `true`/`false`/`unknown`); exit 1 = infrastructure failure (missing `git`/`jq`, not inside a git repository); exit 2 = bad arguments.
+- Emitted key=value fields (stdout + `--github-output` file, using `emit()`'s existing sanitization contract): `would_rotate` (`true`|`false`|`unknown`), `reason` (human-readable string), `age_days` (integer or empty), `age_source` (`pr_created_at`|`branch_commits`|`no-branch`|`unknown`), `accumulated_files` (integer or empty), `accumulated_lines` (integer or empty), `docs_branch`, `rotation_policy` (`none`|`weekly`|`monthly`).
+- New schema field `dossier.ci.thresholds.rotationMaxAccumulatedLines` (integer, minimum 1, default 5000) added to `schema.json`, `settings.json`, and `templates/config.example.json` alongside the existing `thresholds` block — resolvable via the standard `DOSSIER_CI_THRESHOLDS_ROTATION_MAX_ACCUMULATED_LINES` environment override.
+- New workflow step "Check whether the documentation branch would rotate" inside the `policy` job of `dossier-docs-refresh.yml`, placed after the existing `Decide` step, running unconditionally (no `if:` gate on `steps.decide.outputs.should_run`), writing its determination to `$GITHUB_STEP_SUMMARY` via the same append pattern the `Decide` and evidence-bundle steps already use. Executes entirely within the `policy` job's existing `contents: read` / `pull-requests: read` permission scope — no new permission is granted.
+
+## Acceptance Criteria (as validated)
+
+1. Every pipeline run records a clear, human-readable determination of whether branch rotation would currently be warranted, and the specific reason.
+2. No pipeline run, under any circumstance, closes a pull request, deletes a branch, or creates a replacement branch as a result of this capability — it is purely observational.
+3. The determination logic is exercised by tests covering both "would rotate" and "would not rotate" scenarios.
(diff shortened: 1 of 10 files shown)


danielbentes (commented):
## Path A review (paired-reviewer + challenge protocol, 5 facets × skeptic/verifier + holdout ×2)

Self-review (`PR_AUTHOR == CURRENT_USER`). 34 raw findings across 5 facets, 2 auto-consensus pairs, 0 dropped after challenge, 32 consolidated findings. All 32 were triaged and dispositioned this cycle: **14 fixed in-PR** (verified RED→GREEN with regression tests where practical), **4 filed as separate tracked issues** (each requires either a CI-credential-policy decision or touches code outside this PR's diff, not a mechanical fix), **14 left as documented P3 cosmetic/test-coverage notes** on an already-heavily-tested script (81 assertions in `rotation-check.test.sh` alone after this cycle, up from 44).

Two duplicate IDs from independent per-facet enumeration were renamed for uniqueness before posting: the verifier lens's `SEC1`→`SEC1b`, `CONV1`→`CONV1b`, `ERR1`→`ERR1b`, `ERR2`→`ERR2b` (each collided with an unrelated skeptic-lens finding that happened to get the same auto-numbered ID within its own facet).

### Fixed in-PR (14)

| ID | Pri | Facet | Conf | File:line | Summary |
|---|---|---|---|---|---|
| F1 / CONV1b | P2 | code / convention | HIGH | `dossier-rotation-check.sh:188` | `SIZE_THRESHOLD=0` passed the malformed-value guard (0 is all-digits) and made `[ "$ACC_LINES" -ge 0 ]` true unconditionally — the size signal triggered "would rotate" on every run. Live-verified. Guard now also rejects `0`. |
| SEC1 | P2 | security | HIGH | `dossier-rotation-check.sh:275` | `gh pr list --head` matches by branch name only, across every fork with an open PR against this repo — a same-named decoy branch from a fork could spoof the age lookup. Filtered to `isCrossRepository == false`. |
| TEST4 / CONV1 | P2 | test / convention | HIGH | `rotation-check.test.sh:393` | Test 13's `OUT13` was captured but never inspected (SC2034) — the PR's own "shellcheck clean" claim was false at HEAD. Now asserts the raw key=value stdout is untouched by `sanitize_md`. |
| F4 / AgeDaysNegative / ERR3 | P2/P2/P3 | code / test / error-handling | HIGH/MEDIUM/LOW | `dossier-rotation-check.sh:290,361` | A future-dated PR `createdAt` (clock skew) produced a negative `age_days` emitted alongside a reason claiming age is unavailable — self-contradictory, live-verified. `AGE_DAYS` now blanked in the same branch that sets `AGE_KNOWN=0`. |
| TEST3 | P2 | test | HIGH | `dossier-rotation-check.sh:277` | The `gh pr list` failure branch was claimed fixed in an earlier review round but had no regression test. Added (stub `gh` exiting nonzero). |
| TEST8 | P3 | test | LOW | `dossier-rotation-check.sh:290` | No test exercised a future-dated commit/PR at all — closed by the same new test that covers F4/AgeDaysNegative/ERR3. |
| ERR1 | P2 | error-handling | HIGH | `dossier-rotation-check.sh:123` | `cfg()` discarded `$CASCADE`'s exit code, so a genuine config-resolver crash was indistinguishable from a legitimate empty/default read. Live-verified on Ubuntu via a stubbed `mktemp` failure. First fix attempt was itself broken (`die_infra`'s `exit` only killed the `$(...)` subshell) — caught by the regression test before it shipped; fixed by checking the exit status inline in the same shell that calls `cfg()`. |
| SEC1b | P3 | security | HIGH | `dossier-rotation-check.sh:81` | `sanitize_md()` neutralizes table cells (backtick-wrapped) but `note()`'s plain bullets were not wrapped, so a config-derived value could render a spoofed markdown link/heading in the job summary's Notes section. `note()` now backtick-wraps too. |
| UnrecognizedPolicyUntested | P2 | test | HIGH | `dossier-rotation-check.sh:350` | The unrecognized-rotation-policy fallback arm had zero test coverage. Added (also exercises the SEC1b fix above in the same scenario). |
| TEST2 | P2 | test | HIGH (consensus) | `dossier-rotation-check.sh:92` | The `--github-output` newline-injection defense (the only thing preventing a crafted config value from forging extra key=value pairs for a later CI step) was completely untested — the more consequential of the script's two injection surfaces, raised independently by both review lenses. Added. |

### Filed as separate tracked issues (4)

| ID | Pri | Facet | Conf | File:line | Issue | Why not fixed here |
|---|---|---|---|---|---|---|
| SEC2 | P2 | security | MEDIUM | `dossier-policy.sh:390` | [#146](https://github.com/synaptiai/synapti-marketplace/issues/146) | Same unscoped `gh pr list --head` weakness as SEC1, but in pre-existing, already-merged code with privileged (write-token) consequences via the `publish` job's PR-edit path — not touched by this PR's diff. A security fix with its own review, not a telemetry-PR side effect. |
| TEST1 / ERR1b | P2/P2 | test / error-handling | MEDIUM/HIGH | `dossier-rotation-check.sh:199` | [#147](https://github.com/synaptiai/synapti-marketplace/issues/147) | `persist-credentials: false` on the `policy` job's checkout makes `git ls-remote`/`git fetch` fail auth on private repos, silently making the whole feature inert. Fixing this changes CI checkout-credential policy for the job — a bigger design decision than this PR should make unilaterally. |
| TELEMETRY1 | P2 | code | HIGH | `dossier-docs-refresh.yml:219` | [#148](https://github.com/synaptiai/synapti-marketplace/issues/148) | The new step's fields aren't surfaced as job outputs, so the data isn't queryable across runs. Wiring `outputs:` with nothing yet consuming them risks over-engineering; deferred until a real consumer is scoped. |

A fourth issue, [#149](https://github.com/synaptiai/synapti-marketplace/issues/149), was also filed this cycle for the pre-existing `local-merge-hook.test.sh` git-fixture isolation bug flagged in this PR's original body (the root cause — two unchecked `_dossier_safe_mktemp_dir` call sites — was fully diagnosed this cycle, not just symptom-matched as before).

### Documented, not fixed — P3 cosmetic/coverage notes (14)

All narrow, low-impact, or coverage-only on already-correct behavior. Left as-is per this project's cosmetic-P3 convention rather than expanding an already-large fix-forward cycle further.

| ID | Pri | Facet | File:line | Note |
|---|---|---|---|---|
| LOGIC1 | P3 | code | `dossier-rotation-check.sh:81` | `sanitize_md`'s final `tr '\n' ' '` stage is dead code (newline already stripped by the preceding `tr -d`) — cosmetic only, does not reopen the injection risk. |
| LOGIC2 | P3 | code | `dossier-rotation-check.sh:239` | An unresolvable `BASE_REF` (no env var, no `origin/HEAD` symref) is reported as a generic "transport failure" rather than its own accurate reason. Unreachable in CI (the workflow always supplies `BASE_REF`). |
| TEST5 | P3 | test | `dossier-rotation-check.sh:365` | No test pins the exact `>=` threshold boundary for age/size. |
| TEST6 | P3 | test | `dossier-rotation-check.sh:376` | No test exercises both age and size triggering simultaneously (the multi-part REASON string). |
| TEST7 | P3 | test | `dossier-rotation-check.sh:55` | `--help` output isn't pinned by any test. |
| TEST9 | P3 | test | `workflow-template.test.sh:81` | No workflow-template assertion pins the exact `BASE_REF`/`GH_TOKEN` env-expression content. |
| ERR2 | P3 | error-handling | `dossier-rotation-check.sh:125` | Diagnostic notes only ever reach `$SUMMARY_FILE`; a standalone invocation with no `--summary` flag loses all explanatory text (masked in CI, which always passes `--summary`). |
| F2 / ERR2b | P3 | code / error-handling | `dossier-rotation-check.sh:187` | The `SIZE_THRESHOLD` malformed-value guard doesn't call `note()` on fallback, unlike its two sibling guards (`DOCS_BRANCH`/`ROTATION_POLICY`). |
| F3 | P3 | code | `dossier-rotation-check.sh:222` | An invalid `ROTATION_POLICY` value is only flagged in the late Step-3 branch; earlier transport-failure short-circuits report a generic reason instead. Impact bounded (both outcomes are already "unknown"). |
| WeakAgeDaysAssertion | P3 | test | `rotation-check.test.sh:125` | Test 2's age assertion (`-ge 0`) would pass even if the oldest-commit-selection logic picked the wrong commit; verified the current code is correct. |
| DieInfraSummaryUntested | P3 | test | `dossier-rotation-check.sh:108` | `die_infra()`'s `$SUMMARY_FILE`-writing branch has no test coverage. |
| ZeroDiffBranchUntested | P3 | test | `dossier-rotation-check.sh:337` | The genuinely-empty-diff default path (`ACC_FILES`/`ACC_LINES` defaulting to 0) has no direct test; verified manually correct. |
| ERR4 | P3 | error-handling | `dossier-rotation-check.sh:299` | `git rev-list`'s exit code in the commit-walk fallback is never checked; a real git failure and "no Dossier-Generated commit found" produce the same silent empty result. |

### Verification actually performed this cycle

Per this session's own repeated `local-merge-hook.test.sh` git-fixture corruption incidents (root-caused and filed as #149), **the full suite was deliberately not re-run this cycle** — only the targeted files touched by these fixes:

- `bash plugins/dossier/tests/run.sh rotation-check.test.sh workflow-template.test.sh` — 177/177 passing (up from 164 pre-cycle; `rotation-check.test.sh` alone: 81/81, up from 65).
- `shellcheck -S warning` clean on `dossier-rotation-check.sh` and `rotation-check.test.sh`.
- `bash -n` clean on both.
- Every one of the 5 code/behavior fixes (SIZE_THRESHOLD=0, SEC1 fork-scoping, negative AGE_DAYS, ERR1 cfg()-exit-code masking, SEC1b note() markdown neutralization) was reproduced RED against the pre-fix code before the fix, then confirmed GREEN after — including catching a real bug in the first attempt at the ERR1 fix itself (a `die_infra()` call inside a `$(...)` subshell whose `exit` only killed the subshell, not the script).

The PR body's earlier "Full suite: 1802/1802" and "shellcheck clean" verification lines predate this cycle's changes and are being corrected in the PR body to reflect what was actually re-run.

danielbentes:
## Resolution — review cycle 1 (self-review, fix-forward)

All 32 findings from cycle 1's Path A review reached a final disposition this cycle — see the review comment above for the full per-finding breakdown:

- **14 fixed in-PR**, each RED→GREEN verified with a regression test where practical: `TEST2`, `TEST4`/`CONV1` (same underlying SC2034), `SEC1`, `F4`/`AgeDaysNegative`/`ERR3` (same underlying negative-`age_days` bug), `TEST3`, `TEST8`, `ERR1`, `SEC1b`, `UnrecognizedPolicyUntested`, `F1`/`CONV1b` (same underlying `SIZE_THRESHOLD=0` bug).
- **4 filed as separate tracked issues** ([#146](https://github.com/synaptiai/synapti-marketplace/issues/146), [#147](https://github.com/synaptiai/synapti-marketplace/issues/147) ×2, [#148](https://github.com/synaptiai/synapti-marketplace/issues/148)): `SEC2`, `TEST1`/`ERR1b` (same underlying private-repo credential gap), `TELEMETRY1`. Each requires either a CI-credential-policy decision or touches code outside this PR's diff — a deliberate fix, not a mechanical one, so it gets its own issue and review rather than riding along in a telemetry PR.
- **14 documented as P3 cosmetic/coverage notes**, left as-is: `LOGIC1`, `LOGIC2`, `TEST5`, `TEST6`, `TEST7`, `TEST9`, `ERR2`, `F2`/`ERR2b` (same underlying gap), `F3`, `WeakAgeDaysAssertion`, `DieInfraSummaryUntested`, `ZeroDiffBranchUntested`, `ERR4`.

Nothing is open or awaiting a product decision from this cycle — the four filed-as-issue findings are tracked and actionable on their own, not blocked or disputed.

PR #150, opened 2026-08-03
fix(dossier): close mktemp-dir guard exit-code gap, repo-wide

### LINKED ISSUE #149: dossier tests: local-merge-hook.test.sh doesn't check _dossier_safe_mktemp_dir's exit code, reopening the cd "" corruption risk the helper was hardened against
## Summary

`plugins/dossier/tests/lib/assert.sh`'s `_dossier_safe_mktemp_dir()` was hardened (see its own header comment) to `exit 2` on any failure — unset `$RUN_TMPDIR`, a failed `mktemp -d`, or an unusable returned path — specifically so a caller doing `VAR=$(_dossier_safe_mktemp_dir ...)` could never silently receive an empty string. An empty string is dangerous here because `cd ""` **returns exit 0 in bash** (a silent no-op that leaves every subsequent command running in the real caller's directory instead of aborting), even when guarded with `cd "$VAR" || exit 1`.

`plugins/dossier/tests/local-merge-hook.test.sh` has two call sites that don't check the exit code of the command substitution itself:

    FLOWLESS_ROOT=$(_dossier_safe_mktemp_dir "flowless-plugins")   # line 21
    ...
    REPO=$(_dossier_safe_mktemp_dir "local-merge-repo")            # line 32

If `_dossier_safe_mktemp_dir` hits any of its `exit 2` paths, that exit only terminates the subshell spawned by `$(...)` — the assignment still proceeds with `FLOWLESS_ROOT=""` / `REPO=""`, and nothing downstream checks for that. The very next lines then do:

    (
      cd "$REPO" || exit 1
      git init -q -b main
      git config user.email test@example.com
      git config user.name "Test"
      ...
      git commit -q -m "seed"
    ) >/dev/null 2>&1

With `REPO=""`, `cd ""` succeeds, so `|| exit 1` never fires — `git init`, the bare (`--local`-defaulting) `git config user.email`/`user.name`, and later `git checkout -q -b feature-branch main` / `git checkout -q -b conflict-branch main` (lines 78, 123) all run against **the real calling repository**, not an isolated fixture.

## Why this matters here specifically

This is not hypothetical: this exact failure class corrupted this repository's working directory twice during a single session doing unrelated work on issue #138/PR #145 — once creating stray local branches (`feature-branch`, `conflict-branch`, harmless since nothing was pushed), and once overwriting the local `.git/config` `[user]` identity to `Test <test@example.com>`, which briefly misattributed 5 already-committed commits on an unrelated feature branch. Both were diagnosed after the fact via exact fixture-string matching (the `seed`/`feature work`/`conflicting change`/`merge feature-branch` commit messages and the `Test <test@example.com>` author are unique to this file) and repaired non-destructively (`git branch -f` to known-good refs; `git rebase --exec 'git commit --amend --author=...'`). This issue documents the actual mechanism found by re-reading the file afterward, rather than just the symptom.

## Suggested fix

Check the exit code (or emptiness) of both assignments before proceeding, e.g.:

    FLOWLESS_ROOT=$(_dossier_safe_mktemp_dir "flowless-plugins") || exit 2
    [ -n "$FLOWLESS_ROOT" ] || { echo "FATAL: FLOWLESS_ROOT is empty" >&2; exit 2; }

    REPO=$(_dossier_safe_mktemp_dir "local-merge-repo") || exit 2
    [ -n "$REPO" ] || { echo "FATAL: REPO is empty" >&2; exit 2; }

Worth a repo-wide grep for the same unchecked pattern (`VAR=$(_dossier_safe_mktemp_dir ...)` with no immediate exit-code or emptiness check) across the rest of `plugins/dossier/tests/*.test.sh`, since this file is unlikely to be the only caller with the gap — it's just the one two real incidents happened to surface.

danielbentes:
## Summary

Closes #149. `plugins/dossier/tests/lib/assert.sh`'s `_dossier_safe_mktemp_dir()` was hardened to `exit 2` on any internal failure (unset `RUN_TMPDIR`, failed `mktemp -d`, unusable path), specifically so a caller doing `VAR=$(_dossier_safe_mktemp_dir ...)` could never silently receive an empty string — `cd ""` returns 0 in bash, a silent no-op, not an abort. But `exit 2` inside a `$(...)` command substitution only kills that subshell; a caller that doesn't check the assignment's own exit status proceeds with `VAR=""` anyway. `local-merge-hook.test.sh`'s two call sites had exactly this gap, and it corrupted this repo's real working directory twice this session (stray local branches, a briefly-misattributed commit author).

## What changed

- **New**: `_dossier_require_mktemp_dir <varname> <prefix>` in `plugins/dossier/tests/lib/assert.sh` — the sanctioned way to consume `_dossier_safe_mktemp_dir`'s output. Does the capture-and-check exactly once, as a plain function call (never itself wrapped in `$(...)`), so an internal failure's `exit 2` actually propagates through the caller's real shell.
- **Fixed**: `local-merge-hook.test.sh`'s two named call sites (`FLOWLESS_ROOT`, `REPO`).
- **Repo-wide**: the issue flagged this file as "unlikely to be the only caller with the gap." A repo-wide audit confirmed that: 108 more unguarded call sites across 8 other test files, none with any existing exit-code or emptiness guard. All converted to the new helper, including 7 sites in `rotation-check.test.sh` that appended a path suffix to the substitution (e.g. `SUMMARY1=$(...)/summary.md`), split into a two-line dir-then-compose form.
- **Second-order fix** (found by `/flow:pr` review): `setup_fixture()`/`no_gh_path()` in `rotation-check.test.sh` and `setup_fixture()` in `staleness-trigger.test.sh` called the new guard correctly *internally*, but were themselves invoked via `$(...)` at ~24 call sites — reopening the exact bug one level up, since `$(...)` forks a subshell around the whole function body too. Converted both to the same out-parameter convention (`printf -v` into a caller-named variable, called as a plain statement) as the guard itself, and updated every call site.
- **Third-order fix** (found by `/flow:review`'s paired-reviewer round): the wrapper functions' own out-parameter handoff was itself unguarded — the same "`printf -v` can fail silently" class already fixed in the main helper had been reintroduced one level up, in the second-order fix's own new code. Extracted a shared `_dossier_assign_outvar <varname> <value>` helper (guarded, matching the FATAL+exit-2 pattern) and had every out-parameter assignment in the file delegate to it. Also added a static lint (`mktemp-guard.test.sh` scenario 6, pure awk/grep) that permanently guards against any future helper reintroducing this whole bug class, `local`-scoped `no_gh_path`/`setup_fixture`'s internal variables (previously inert only by accident of always running in a subshell — now actually enforced), and added `local _dir` to `_dossier_safe_mktemp_dir` as proactive hardening.

## Review

Two rounds. `/flow:pr`'s 5-agent parallel review (code-quality, convention, security, test-runner, error-handling): convention and test-runner clean; code-quality, security, and error-handling independently converged on the second-order guard-swallow from three different angles, fixed and re-verified. A reviewer's `local _dir` suggestion was investigated and empirically disproven as a *live-bug* fix at the time (the guard's own internal `$(...)` call already isolates `_dir`) — not applied in that round.

`/flow:review`'s Path A paired-reviewer round (10 agents, 5 facets × skeptic/verifier) re-reviewed the PR at that HEAD and found 14 more findings, converging across 2-4 independent reviewers each on: the wrapper out-parameter handoff being unguarded, `no_gh_path`/`setup_fixture` needing `local` scoping, no durable lint against reintroduction, and `local _dir` re-raised with a different framing (not "fixes a live bug" but "the file's own docstring actively directs future contributors toward a refactor that would make it live") — accepted this time as cheap, zero-behavior-change proactive hardening. 8 findings fixed in-PR, 1 filed as a separate tracked issue ([#151](https://github.com/synaptiai/synapti-marketplace/issues/151) — an unrelated, pre-existing vacuous-pass bug in `disclosure-gate.test.sh`, not touched by this PR), 5 documented as reasoned not-fixed. Full per-finding breakdown in the [review comment](https://github.com/synaptiai/synapti-marketplace/pull/150#pullrequestreview-4843425289) and [resolution comment](https://github.com/synaptiai/synapti-marketplace/pull/150#issuecomment-5165619674).

## Verification

- [x] Full suite: `bash plugins/dossier/tests/run.sh` — **1852/1852** (1798 pre-change → 1844 after round 1 → 1848 after round 2 → 1852 after round 3)
- [x] `shellcheck -S warning -x plugins/dossier/tests/*.sh` (exact CI invocation) — clean, exit 0
- [x] `bash -n` on every modified file — clean
- [x] Regression test (`mktemp-guard.test.sh`) reproduces the exact corruption mechanism RED against the pre-fix `local-merge-hook.test.sh`, then confirms GREEN — plus pattern-level regression tests for the second-order (scenario 4) and third-order (scenario 5) fixes, independent of any specific file
- [x] Static lint (`mktemp-guard.test.sh` scenario 6): no `*.test.sh` file defines a guard-consuming function that's also invoked via `$(...)`/backticks anywhere in that file — verified it actually catches a deliberately-introduced violation before trusting it as a permanent regression guard
- [x] CI: 4/4 checks green (`test`, `CodeQL`, `Analyze (actions)`, `Analyze (python)`), independently re-checked via `gh pr checks` per this project's own memory of prior local-green/CI-red divergence on this exact plugin
- [x] TDD throughout: every fix reproduced RED against the pre-fix code before being confirmed GREEN

Full detail (specification, acceptance criteria, verification evidence, all three rounds' findings) in `.decisions/issue-149.md`.

### DIFF
--- a/.decisions/issue-149.md
+++ b/.decisions/issue-149.md
@@ -34,13 +34,23 @@ created: '2026-08-03T10:03:20Z'
 
 PASS — 3 tasks reviewed. Task "add guarded helper" specifies the exact function signature, its two internal guards, and the printf -v assignment mechanism. Task "fix local-merge-hook.test.sh" names the exact two call sites (file:line) and the exact verification command. Task "repo-wide conversion" specifies the exact audit method (Python scan, zero pre-existing guards confirmed) and the two conversion shapes (simple swap; two-line split for the 7 path-suffix sites in rotation-check.test.sh).
 
+## Second-order finding (PR #150 review, fix-forwarded before merge)
+
+The first-pass fix (direct call-site conversion) left one gap: `setup_fixture()`/`no_gh_path()` in `rotation-check.test.sh` and `setup_fixture()` in `staleness-trigger.test.sh` called the new guard correctly *internally*, but were themselves invoked via `$(...)` at ~24 call sites — reopening the exact swallowed-exit bug one level up, since command substitution forks a subshell around the whole function body too. Found independently by three `/flow:pr` review agents (error-handling, security, code-quality) from three different angles. Fixed by converting both functions to the same out-parameter convention as the guard itself (`printf -v` into a caller-named variable, called as a plain statement), updating every call site, and hardening the guard's own final `printf -v` line to match. A repo-wide re-audit (every `$(...)`-invoked function name cross-checked against every function body containing the guard, across all 10 test files) confirmed zero remaining instances of this pattern.
+
+One reviewer's suggested prerequisite (`local _dir` in `_dossier_safe_mktemp_dir`, to prevent a claimed loop-variable clobber in `no_gh_path`) was investigated and empirically disproven via a standalone repro script — the guard's own internal `$(_dossier_safe_mktemp_dir ...)` call already forks its own subshell, which isolates `_dir` regardless of whether the outer helper is itself wrapped in `$(...)`. Not applied.
+
+Also checked: converting `setup_fixture`/`no_gh_path` from subshell-isolated to plain-statement execution means their previously-unscoped internal variables (`_bare`, `_clone`, `_out`, `_dir`, `_entry`, `_base`, `OLD_IFS`) now persist in the caller's real shell instead of dying with a subshell. Grepped for reads of all seven names outside their three defining functions (`no_gh_path`, `setup_fixture`, `push_docs_branch_commit`) in `rotation-check.test.sh` — zero external reads found, so the newly-persistent globals are inert (each function reassigns before reading, so there's no cross-call contamination either).
+
 ## Verification
 
-- Full suite: `bash plugins/dossier/tests/run.sh` — 1844/1844 (up from 1798 pre-change)
+- Full suite: `bash plugins/dossier/tests/run.sh` — 1848/1848 (up from 1798 pre-change, then 1844 after the first fix pass, then 1848 after the second-order fix's new mktemp-guard.test.sh scenario 4)
 - `shellcheck -S warning -x plugins/dossier/tests/*.sh` (exact CI invocation) — clean, exit 0
 - `bash -n` on every modified file — clean
 - Regression test (`mktemp-guard.test.sh` scenario 3) reproduced RED against the pre-fix `local-merge-hook.test.sh` (confirmed a `.git` directory materializes in an isolated scratch fixture when the guard doesn't fire), then confirmed GREEN after the fix
-- Repo-wide audit: 0 remaining `VAR=$(_dossier_safe_mktemp_dir ...)` call sites without a guard (116 total call sites found; 116 now guarded)
+- Regression test (`mktemp-guard.test.sh` scenario 4) regression-tests the second-order pattern generically (stdout-returning wrapper vs. out-parameter wrapper), independent of any specific file
+- Repo-wide audit: 0 remaining `VAR=$(_dossier_safe_mktemp_dir ...)` direct call sites without a guard (116 total direct call sites found; 116 guarded)
+- Repo-wide re-audit (second-order): 0 remaining functions that call the guard internally and are invoked via `$(...)` by their own callers
 
 <!-- auto-log: 2026-08-03 Write /Users/danielbentes/synapti-marketplace/.decisions/issue-149.md -->
 
@@ -63,3 +73,9 @@ PASS — 3 tasks reviewed. Task "add guarded helper" specifies the exact functio
 <!-- auto-log: 2026-08-03 12:33 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/lib/assert.sh -->
 
 <!-- auto-log: 2026-08-03 12:39 Edit /Users/danielbentes/synapti-marketplace/plugins/dossier/tests/lib/assert.sh -->
+
+<!-- auto-log: 2026-08-03 12:41 commit "fix(dossier): close second-order guard-swallow in setup_fixture/no_gh_path" -->
+
+<!-- auto-log: 2026-08-03 12:42 Write /private/tmp/claude-501/-Users-danielbentes-synapti-marketplace/8c76ed85-3c0c-4a32-8924-be0cf2c7bc2d/scratchpad/pr149-body.md -->
+
+<!-- auto-log: 2026-08-03 12:46 Edit /Users/danielbentes/synapti-marketplace/.decisions/issue-149.md -->
--- a/.decisions/issue-149.md
+++ b/.decisions/issue-149.md
@@ -1,0 +1,47 @@
+---
+issue: 149
+created: '2026-08-03T10:03:20Z'
+---
+# Issue #149 — mktemp-dir guard: exit-code check reopens the `cd ""` corruption risk
+
+**Title**: dossier tests: local-merge-hook.test.sh doesn't check `_dossier_safe_mktemp_dir`'s exit code, reopening the `cd ""` corruption risk the helper was hardened against
+**Labels**: bug, plugin
+
+## Specification
+
+### Non-goals
+
+- Not touching the two real, already-corrected corruption incidents from earlier this session (stray local branches, misattributed commit authors) — those are already repaired; this issue is about closing the mechanism that caused them, not re-litigating the incident.
+- Not changing `_dossier_safe_mktemp_dir`'s own internal behavior (its `exit 2`-on-failure contract is correct and intentional) — the fix is entirely at the call-site layer.
+
+### Failure modes
+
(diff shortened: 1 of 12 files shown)


danielbentes (commented):
## Self-review: PR #150 (issue #149)

Path A paired-reviewer round (10 agents: code-quality/security/convention/test-runner/error-handling × skeptic+verifier) reviewed the PR at its round-2 HEAD (after the second-order guard-swallow fix). This is a self-review — every P1/P2 finding was fixed in-PR before this comment was posted; see the resolution comment for the RESOLVED marker.

### Fixed in-PR (14)

| ID | Priority | Category | Location | Disposition |
|---|---|---|---|---|
| LOCAL1 | P2 | correctness | rotation-check.test.sh:47,81 | consensus (3 reviewers) — `no_gh_path`/`setup_fixture`'s internal vars weren't `local`; scoped them |
| GUARD1 | P2 | error-handling | rotation-check.test.sh:67,97; staleness-trigger.test.sh:70 | consensus (2 reviewers) — the wrapper functions' own out-parameter `printf -v` was itself unguarded; extracted a shared `_dossier_assign_outvar` helper |
| LINT1 | P2 | maintainability | mktemp-guard.test.sh | consensus (3 reviewers) — no durable guard against a future helper reintroducing this bug class; added a static lint (scenario 6) |
| SCEN3 | P2 | testing | mktemp-guard.test.sh:44-50 | unchallenged — scenario 3's `.git`-absence assertion was proven vacuously (the first of two guards always fires before the second); relabeled to describe what's actually proven |
| CIPEND1 | P2 | convention | PR CI status | unchallenged — CI's `test` check was still pending when first checked; re-confirmed green before proceeding |
| DIRLOCAL1 | P3 | security | assert.sh:43 | consensus (2 reviewers) — `_dossier_safe_mktemp_dir`'s internal `_dir` wasn't `local`; added as proactive hardening (not a live bug under the current call graph) |
| RCASSERT1 | P3 | testing | mktemp-guard.test.sh:69-74 | unchallenged — the broken-wrapper scenario captured `$?` but never asserted it was actually `2`; added the assertion |
| LABELS1 | P3 | convention | PR metadata | consensus (2 reviewers) — PR had no labels though the linked issue specifies `bug, plugin`; synced |

### Filed as a separate issue (1)

| ID | Priority | Category | Location | Issue |
|---|---|---|---|---|
| DISCLOSURE1 | P3 | testing | disclosure-gate.test.sh:288-303 (unrelated file, not touched by this PR) | [#151](https://github.com/synaptiai/synapti-marketplace/issues/151) — a fixture write path bug makes an unrelated assertion pass vacuously |

### Documented, not fixed (5)

| ID | Priority | Category | Location | Reasoning |
|---|---|---|---|---|
| OUTVARCOLL1 | P3 | security | assert.sh:77-88 | `_dossier_require_mktemp_dir`'s own internal locals (`__dossier_mktemp_varname` etc.) could theoretically collide with a caller-supplied outvar name — but they're deliberately namespaced specifically to make that vanishingly unlikely, no current call site is close, and adding entry validation for a "won't happen by construction" case is exactly the over-engineering CLAUDE.md's guidance warns against |
| NAMING1 | P3 | convention | assert.sh:54 vs :72; rotation-check.test.sh:53 | the shared helper's docstring names its own signature `<varname>` while telling future authors to use `<outvar>` (every derived function does); a local-declaration placement inconsistency in `no_gh_path`. Both cosmetic, no functional risk, shellcheck clean either way |
| DIAGTEXT1 | P3 | testing | mktemp-guard.test.sh:28-29,87-88 | scenarios 2 and 4b assert exit code + marker-absence but not the FATAL diagnostic text itself; confirmed the text is genuinely emitted today, so the current pass is for the right reason — marginal value to add given the exit-code+marker pair already fully pins the behavior |
| ERR3PRE | P3 | error-handling | local-merge-hook.test.sh:44 (`run_hook`), staleness-trigger.test.sh:73 (`run_policy`) | both predate this PR, don't call the guard today, and match the vulnerable shape closed elsewhere (stdout-returning, invoked via `$(...)`) — out of scope for this PR since neither file's relevant lines are touched by this diff; worth a comment if either function ever grows a guard call |
| SHADOW1 | P3 | error-handling | rotation-check.test.sh, staleness-trigger.test.sh | the out-parameter convention is vulnerable to a caller passing an outvar name that shadows one of the wrapper's own locals; not reachable today (no call site's outvar name collides), addressed via a docstring warning rather than a rename (renaming risks introducing a real bug while fixing a latent, unreached one) |

danielbentes:
## Self-review resolution: PR #150 (cycle 1)

All 14 findings from the paired-reviewer round were dispositioned: 8 fixed in-PR (RED→GREEN verified where applicable, full suite re-confirmed at 1852/1852 after each fix), 1 filed as a separate tracked issue ([#151](https://github.com/synaptiai/synapti-marketplace/issues/151), unrelated file not touched by this PR), 5 documented as reasoned not-fixed (each with an explicit reason in the review comment — mostly "unreached given the current call graph, fixing would add unjustified risk/complexity for the value").

Per this project's finding-ledger schema, every finding ID appears in `RESOLVED` regardless of disposition — the full reasoning for each lives in the review comment above, not in this marker.

PR #155, opened 2026-08-03
fix(dossier): authenticate rotation-check's git network calls on private repos

### LINKED ISSUE #147: dossier: policy job's persist-credentials: false makes rotation-check (and any git network op) silently inert on private repos
## Summary

The `policy` job in `dossier-docs-refresh.yml` checks out with `persist-credentials: false` (by design — it only needs `contents: read`/`pull-requests: read`, and the job never pushes). `dossier-rotation-check.sh` (added in #138/PR #145) runs `git ls-remote`/`git fetch` against the `origin` remote as unauthenticated HTTPS. On a **public** repo this works fine. On a **private** repo, these git network operations will fail authentication — the script's own transport-failure handling degrades this to `age_source=unknown`/`would_rotate=unknown` rather than crashing, so the symptom is silent: the feature just never produces a real determination, and nothing surfaces this as a configuration problem.

Flagged during the Path A review of PR #145 (verifier lens, error-handling facet). Not fixed in that PR — deliberately scoped out, since the fix changes CI checkout-credential policy for the `policy` job, which is a bigger design decision than a telemetry-only PR should make unilaterally.

## Where this lives

- `plugins/dossier/templates/ci/dossier-docs-refresh.yml` — `policy` job's checkout step (`persist-credentials: false`)
- `plugins/dossier/bin/dossier-rotation-check.sh` — Step 1/Step 2 (`git ls-remote --exit-code`, `git fetch`)

## Options to consider

1. Authenticate git operations via the job's existing `GH_TOKEN` (e.g. `git -c http.extraheader="AUTHORIZATION: bearer $GH_TOKEN" ls-remote ...`, or `gh auth setup-git`/a credential helper) — narrower than flipping `persist-credentials` for the whole job.
2. Detect the auth failure specifically (vs. a generic transport failure) and emit a note that explicitly names "private repository, no checkout credentials" as the reason, rather than a generic "transport failure."
3. Accept the limitation and document it prominently in the script's usage header and this repo's dossier docs.

Needs a decision on which approach fits the project's threat model for this job before implementing — this is a CI credential-policy change, not a simple bug fix.
## Acceptance Criteria

- [ ] `dossier-rotation-check.sh`'s `git ls-remote` and both `git fetch` calls (lines 223, 267, 269) authenticate against `origin` using `GH_TOKEN` when it is non-empty, via a scoped `-c http.extraheader="AUTHORIZATION: bearer $GH_TOKEN"` applied per-invocation — not `persist-credentials: true`, not `gh auth setup-git`, and not a persistent credential helper.
- [ ] When `GH_TOKEN` is empty or absent, behavior is byte-for-byte unchanged from today: unauthenticated git calls, working on public repos, degrading to `age_source=unknown`/`would_rotate=unknown` on private repos without a token. No regression to the public-repo path.
- [ ] The auth logic lives inside `dossier-rotation-check.sh` itself, not the calling workflow YAML — the script remains self-contained and correct wherever invoked.
- [ ] New test coverage (extending `plugins/dossier/tests/rotation-check.test.sh`, which already owns this script's bare-repo test harness) using a PATH-stub `git` wrapper that logs its own argv then execs the real `git`, asserting the auth header is present in the `ls-remote`/`fetch` invocations when `GH_TOKEN` is set, and absent when it is empty.
- [ ] The script's usage header comment (currently: "GH_TOKEN passed through to `gh`, never interpolated into a command") is updated to document that GH_TOKEN now also authenticates the script's own git network calls.

## Decision on approach (resolved via direct interview before implementation)

Of the three options this issue lists, **option 1** (authenticate via the existing `GH_TOKEN`) is the approach — confirmed empirically that `GH_TOKEN: ${{ github.token }}` is already present, unused for this purpose, in the exact workflow step that invokes this script (`dossier-docs-refresh.yml:239`), and that the `policy` job's permissions are already minimal (`contents: read`, `pull-requests: read`), so no privilege escalation is introduced. Full reasoning will be recorded in `.decisions/issue-147.md`.

danielbentes:
## Summary

Closes #147. `dossier-rotation-check.sh`'s `git ls-remote`/`git fetch` calls against `origin` ran as unauthenticated HTTPS — working on public repos, silently degrading to `age_source=unknown` on private repos rather than surfacing a configuration problem. `GH_TOKEN` was already threaded into the workflow step that invokes this script (used implicitly by the script's own `gh pr list` call), just never used for git's own network calls.

Design decided via a direct pre-implementation interview (three rounds, six sub-decisions — see `.decisions/issue-147.md`): authenticate via the existing `GH_TOKEN`, inside the script itself (not the workflow YAML), with a PATH-stub test technique, falling back to today's behavior when `GH_TOKEN` is empty/absent.

## What shipped (materially different from the original plan — see below)

- New `git_auth()` wrapper in `dossier-rotation-check.sh`, used at all three network call sites (`ls-remote`, two `fetch` calls).
- Credential passed via `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_0`/`GIT_CONFIG_VALUE_0` environment variables (git ≥2.31), scoped to a single subprocess via the `VAR=val cmd` prefix form — not exported globally, not a `-c` argv flag.
- **Basic auth, not Bearer**: verified live against a real private GitHub repo that the issue's own suggested `AUTHORIZATION: bearer $GH_TOKEN` format is rejected outright; `AUTHORIZATION: basic base64(x-access-token:$GH_TOKEN)` (the same shape `actions/checkout` constructs internally) succeeds.
- **Scoped to origin's own host**, not a bare unscoped key: a code-review pass live-reproduced that an unscoped `http.extraheader` key sends the credential to *any* HTTPS host the git invocation touches, not just `origin` — confirmed by sending a GitHub-derived credential to gitlab.com. Fixed by deriving `origin`'s scheme+host via `git remote get-url origin` and scoping the key accordingly (matching `actions/checkout`'s own narrower scoping). Non-HTTP(S) origins fall through unauthenticated entirely.
- `GH_TOKEN` and the base64 output are control-character-stripped before use (matching this file's existing `sanitize_md()`/`emit()` convention).

## Review process

Two review passes, both fix-forward before this PR reached its final state:

1. **Self-review** (`Agent(code-reviewer)`) on the initial implementation: found and fixed a stale `--help` range, moved the auth header off `-c` argv (process-listing exposure) onto env vars, and — critically — a manual verification against a real private repo (prompted by the reviewer correctly noting the local test fixture cannot exercise real HTTP auth at all) caught that the issue's own suggested Bearer format doesn't work.
2. **Parallel 5-agent review** (code-quality, convention, security, error-handling, test-runner) on the resulting branch: found and fixed the P1 unscoped-header issue above, plus several P2/P3s (unchecked base64-pipeline exit status, non-`local` credential variables, missing GH_TOKEN-set diagnostics in failure notes, a stale goal-file specification, and a test-authoring bug where a `$(...)`-based dirty token never actually carried a surviving embedded newline).

Full narrative, every finding, and the live-verification transcripts are in `.decisions/issue-147.md`.

## Testing

- `bash plugins/dossier/tests/run.sh rotation-check.test.sh` — 96/96 (up from 74 pre-PR), including new scenario 21C (non-HTTP(S) origin regression test) and scoping assertions in 21A/22.
- `shellcheck -S warning -x` on both touched files — clean.
- Manual verification against a real private GitHub repo (read-only `git ls-remote`, non-mutating): confirmed the shipped `git_auth()` function, extracted verbatim from the file, authenticates correctly against `origin` and does not send the credential to an explicitly different host in the same session.
- Full local suite currently unreliable due to a local disk-space issue (documented mid-session, unrelated to this change) — relying on GitHub Actions CI for full-suite confirmation, consistent with this plugin's own prior guidance to verify actual CI rather than local-only runs.

## Known residual limitation (documented, not blocking)

The live verification used a personal access token (`gh auth token`), not GitHub Actions' `github.token` (the actual credential this script receives in production) — that token only exists for the duration of a real job. Both are documented to use the same `x-access-token` Basic-auth convention. If a real CI run against a private repo ever shows different behavior, that's the next signal to act on.

### DIFF
--- a/.decisions/issue-147.md
+++ b/.decisions/issue-147.md
@@ -352,3 +352,5 @@ rotation-check.test.sh` — 96/96. `shellcheck -S warning -x` — clean.
 <!-- auto-log: 2026-08-03 19:51 Edit /Users/danielbentes/synapti-marketplace/.flow/goals/issue-147.goal.yaml -->
 
 <!-- auto-log: 2026-08-03 19:53 commit "fix(dossier): scope git_auth's credential to origin's own host, fix silent-failure edge cases" -->
+
+<!-- auto-log: 2026-08-03 19:54 Write /private/tmp/claude-501/-Users-danielbentes-synapti-marketplace/8c76ed85-3c0c-4a32-8924-be0cf2c7bc2d/scratchpad/pr-147-body.md -->
--- a/.decisions/issue-147.md
+++ b/.decisions/issue-147.md
@@ -1,0 +1,241 @@
+---
+issue: 147
+created: '2026-08-03T16:00:00Z'
+artifacts:
+- type: specification
+  captured_at: '2026-08-03T16:00:11Z'
+  by: direct-interview
+  elements:
+  - non-goals
+  - failure-modes
+  - interface-contracts
+- type: goal-created
+  captured_at: '2026-08-03T16:00:11Z'
+  goal_id: issue-147
+  source: github_issue:147
+- type: workflow-run
+  captured_at: '2026-08-03T16:00:33Z'
+  workflow: start-issue
+  run_id: 2026-08-03T160000Z-issue-147
+  status: active
+---
+# Issue #147 — dossier-rotation-check.sh: git network calls are silently inert on private repos
+
+**Title**: dossier: policy job's persist-credentials: false makes rotation-check (and any git network op) silently inert on private repos
+**Labels**: bug, plugin
+
+## Specification
+
+_Captured directly from a pre-implementation interview with the user (three rounds, six
+sub-decisions total), each grounded in direct codebase verification rather than the issue
+body's claims alone. See below for what was verified._
+
+### Non-goals
+
+- Not changing `persist-credentials: false` on the `policy` job's checkout step, and not
+  flipping it to `true` for the whole job. The fix is scoped to authenticating three specific
+  git invocations inside `dossier-rotation-check.sh`, not broadening the job's credential
+  footprint.
+- Not using `gh auth setup-git` — confirmed during the interview that this would install a
+  persistent git credential helper for the rest of that step's process, a broader footprint
+  than the 2-3 calls that actually need authentication, and one that would need independent
+  duplication at any future call site rather than being self-contained to this script.
+- Not treating an empty/absent `GH_TOKEN` as a hard configuration error. The script must remain
+  usable in contexts where `GH_TOKEN` is legitimately unset (local runs, a future caller that
+  doesn't export it, public repos where it was never needed) — falling back to today's
+  unauthenticated behavior, never regressing the working public-repo case.
+- Not building new test infrastructure (an HTTP server, a git daemon) to exercise the
+  authenticated request end-to-end. The existing `rotation-check.test.sh` harness uses a local
+  filesystem bare-repo remote, against which an HTTP-only `extraheader` is inert (git ignores
+  `http.*` config for non-HTTP transports) — verified this is harmless, not a blocker, and
+  scoped testing to argv-inspection instead (see Interface contracts).
+
+### Failure modes
+
+- **Timeouts** — none — no new network call is introduced; the existing `git ls-remote`/
+  `git fetch` calls are the same calls, now carrying an additional header when authenticated.
+- **Partial failures** — none — the fix is confined to how the git commands are invoked
+  (an added `-c` flag), not a multi-step operation that could partially complete.
+- **Invalid input** — `GH_TOKEN` containing shell metacharacters must never be interpolated into
+  a command string that reaches a shell (matching the script's existing documented principle,
+  header line 15: "GH_TOKEN passed through to `gh`, never interpolated into a command"). **Note
+  on the shipped shape**: the original plan below (a `-c http.extraheader=...` argv value) was
+  superseded during self-review by `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_0`/`GIT_CONFIG_VALUE_0`
+  environment variables instead (process-listing exposure, see Self-review fix-forward below) —
+  the "never reaches a shell" property holds identically either way, since env vars set via the
+  `VAR=val cmd` prefix form are not shell-interpolated any more than an argv entry is.
+- **Missing context** — `GH_TOKEN` empty or absent: falls back to today's unauthenticated
+  behavior (see Non-goals above) rather than failing loudly, since a public-repo caller that
+  never needed a token must not regress.
+
+### Interface contracts
+
+- **Superseded during self-review** (final shipped shape — see below): `dossier-rotation-check.sh`'s
+  `git_auth()` wrapper conditionally sets `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=http.extraheader
+  GIT_CONFIG_VALUE_0="AUTHORIZATION: basic base64(x-access-token:$GH_TOKEN)"` (not the `-c
+  ...bearer...` form originally planned here) before every `git ls-remote --exit-code --heads
+  origin "$DOCS_BRANCH"` (line 223) and `git fetch --no-tags [--prune] origin ...` call (lines
+  267, 269), added only when `GH_TOKEN` is non-empty.
+- No change to the script's `--github-output`/`--summary` output contract, its documented exit
+  codes, or any of its 8 emitted fields (`would_rotate`/`reason`/`age_days`/`age_source`/
+  `accumulated_files`/`accumulated_lines`/`docs_branch`/`rotation_policy`).
+- No change to `dossier-docs-refresh.yml` — `GH_TOKEN: ${{ github.token }}` is already present
+  in the "Check whether the documentation branch would rotate" step's `env:` block
+  (confirmed: line 239), so no new wiring is needed there.
+- Test additions land in `plugins/dossier/tests/rotation-check.test.sh`, using a PATH-stub `git`
+  wrapper that logs its own argv to a file then execs the real `git` binary found via
+  `command -v` before the stub is placed on `PATH` — preserving the existing bare-repo
+  fixture's real git behavior while making the constructed command line assertable. Scoped
+  narrowly: the stub sits ahead of real `git` on `PATH` only for the invocation of
+  `dossier-rotation-check.sh` under test, not for the surrounding fixture-setup code (which
+  needs real git to build the bare repo and push branches).
+- The script's usage header (currently line 15: "GH_TOKEN passed through to `gh`, never
+  interpolated into a command") is revised to state GH_TOKEN now also authenticates the
+  script's own `git` network calls, not just `gh`.
+
+## Decision on approach
+
+Of the three options the issue itself named:
+
+1. Authenticate via the existing `GH_TOKEN` — **selected**.
+2. Better diagnostics only (name "private repo, no credentials" instead of a generic transport
+   failure), leaving the feature permanently inert on private repos — rejected: doesn't fix the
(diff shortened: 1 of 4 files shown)


danielbentes (commented):
## Self-Review Summary

Performed self-review with fix-forward. Two rounds: (1) a single-agent self-review immediately after implementation, (2) a 5-agent parallel dispatch (code-quality, convention, security, error-handling, test-runner) as part of `/flow:pr`'s own review phase. Both rounds are combined into this one cycle since neither produced a separate posted review before this one. Path B (single-session dispatch, no paired-reviewer challenge round) — marker uses the legacy 5-field form.

### Verdict Summary

Verdict: PASS (4/5 criteria via independent verdict-judge; AC1 was NEEDS-HUMAN-REVIEW — the local test fixture cannot exercise real HTTP auth, so the only evidence for "does the header actually authenticate" was a manual verification against a real private repo, which the judge's independence protocol correctly excludes as inadmissible. Approved after review — see `.decisions/issue-147.md`.)

### Findings Found & Fixed

| Finding | Fix Applied |
|---------|-------------|
| **F1 · P1 · `plugins/dossier/bin/dossier-rotation-check.sh:232`**<br>`GIT_CONFIG_KEY_0=http.extraheader` was unscoped — applies to every HTTPS host the git invocation touches, not just `origin`. Live-reproduced: sent a GitHub-derived credential to gitlab.com. | Scoped to `origin`'s own resolved scheme+host (`http.<scheme>://<host>/.extraheader`), matching `actions/checkout`'s own narrower scoping. Non-HTTP(S) origins fall through unauthenticated entirely. Re-verified live: the credential no longer reaches an explicitly different host. |
| **F2 · P2 · `.flow/goals/issue-147.goal.yaml`**<br>AC1's text and `interface_contracts` still described the superseded `-c http.extraheader=...bearer...` design; AC3's verification command spuriously passed against the shipped code since the substring survived. | Updated to describe the shipped `GIT_CONFIG_*`/Basic-auth/scoped-host mechanism. |
| **CONV1 · P2 · `plugins/dossier/tests/rotation-check.test.sh` (scenario 22)**<br>The dirty-token test built its CRLF via `"...$(printf '\r\n')"` — a `$(...)` substitution strips its own trailing newline regardless of concatenation position, so the token never actually carried a surviving embedded newline, contradicting the test's own comment and commit message. | Rebuilt via ANSI-C quoting (`$'...\r\n...'`), which embeds real control characters with no such stripping. |
| **ERR1 · P2 · `plugins/dossier/bin/dossier-rotation-check.sh:349,377`**<br>A rejected/invalid `GH_TOKEN` produces exit 128 — worse than the unauthenticated case for a public repo that would otherwise work. Live-reproduced against a real public repo. | Not fixed via retry-without-auth (considered and rejected: masks a genuine "your token lacks access" signal, doubles network calls, and the real production credential — GitHub Actions' `github.token` — is always valid for the job's duration, so this is a local-invocation edge case, not a CI-path regression). Addressed via diagnostics: both transport-failure notes now record `GH_TOKEN set: yes/no`, so a rejected token is distinguishable from a generic network problem. |
| **ERR2 · P2 · `plugins/dossier/bin/dossier-rotation-check.sh:221`**<br>The base64-encoding pipeline's exit status was never checked; a broken encoder would silently produce an empty credential, sent as a malformed header and rejected — worse than no credential at all. | Falls through unauthenticated when the encoding result is empty, rather than sending a header known to be broken. |
| **ERR3 · P2 · `plugins/dossier/bin/dossier-rotation-check.sh:231`**<br>`GIT_CONFIG_COUNT=1` unconditionally claims the whole `GIT_CONFIG_*` namespace, silently overriding any ambient entries a caller might have set. Nothing in this repo sets this today (confirmed via grep) — a latent gap, not a live one. | Documented explicitly in `git_auth()`'s own comment as an intentional exclusive claim, rather than fixed with speculative append-logic for a scenario with no current reachable caller. |
| **ERR4 · P3 · `plugins/dossier/bin/dossier-rotation-check.sh:349,377`**<br>Transport-failure notes didn't record whether `GH_TOKEN` was set, collapsing three distinguishable causes into identical wording. | Both notes now include `GH_TOKEN set: yes/no`. |
| **SEC1/ERR5 · P3 · `plugins/dossier/bin/dossier-rotation-check.sh:183`**<br>`_clean_token`/`_b64_creds` were plain global assignments (raised independently by two agents), needlessly widening the credential's exposure window in the shell's namespace after the function returns. | Added `local`. |
| **SEC2 · P3 · `plugins/dossier/bin/dossier-rotation-check.sh:167`**<br>`git_auth()`'s docstring didn't state that callers must redirect stdout+stderr — the base64-transformed credential isn't GitHub Actions' literal secret string, so its log masking wouldn't catch a leak from a future call site that omits the redirect. | Added the caller contract to the docstring. |

None — all findings fixed in-PR. AC1's NEEDS-HUMAN-REVIEW verdict was resolved via direct user review of the manual private-repo verification, not deferred as an open item.

### Verification

- [x] Quality commands pass after fixes — `bash plugins/dossier/tests/run.sh rotation-check.test.sh`: 96/96. `shellcheck -S warning -x`: clean.
- [x] No new issues introduced — the P1 fix was re-verified against the real private repo with the exact shipped function (not a hand-written replica), both for successful auth against `origin` and for non-leakage to an explicitly different host.
- [x] All P1/P2/P3 findings in touched files fixed in-PR or escalated

danielbentes:

### Review Cycle Summary

| Metric | Count |
|--------|-------|
| Received | 9 |
| Fixed | 9 |
| Discussed | 0 |
| Escalated | 0 |

### Changes Made

**1. F1 — unscoped auth-header credential leak (P1)**
- Scoped `GIT_CONFIG_KEY_0` to `origin`'s own resolved scheme+host instead of a bare `http.extraheader` key; non-HTTP(S) origins fall through unauthenticated. Re-verified live against a real private repo (still authenticates) and a real third-party host (no longer leaks).
- Commit: `edf7186`

**2. F2/SEC3 — stale goal-file specification (P2)**
- Updated `.flow/goals/issue-147.goal.yaml`'s AC1 text and `interface_contracts` to describe the shipped mechanism.
- Commit: `821431e`

**3. CONV1 — test CRLF-embedding bug (P2)**
- Rebuilt scenario 22's dirty token via ANSI-C quoting so the embedded CRLF actually survives.
- Commit: `edf7186`

**4. ERR1/ERR2/ERR3/ERR4 — error-handling gaps (P2/P3)**
- Diagnostic improvement for GH_TOKEN-rejected vs. absent (ERR1/ERR4); fall through unauthenticated on empty encoding result (ERR2); documented exclusive `GIT_CONFIG_*` namespace claim (ERR3).
- Commit: `edf7186`

**5. SEC1/ERR5/SEC2 — credential hygiene (P3)**
- `local`-scoped credential variables; documented the caller stdout/stderr-redirect contract.
- Commit: `edf7186`

### Discussion Points

None.

### Escalated for Human Judgment (if any)

None.

### Thread Status

| Thread | Status |
|--------|--------|
| **F1 — unscoped credential leak** | Resolved |
| **F2/SEC3 — stale goal-file spec** | Resolved |
| **CONV1 — test CRLF bug** | Resolved |
| **ERR1/ERR2/ERR3/ERR4 — error-handling gaps** | Resolved |
| **SEC1/ERR5/SEC2 — credential hygiene** | Resolved |

### Verification

- [x] All quality checks pass — `bash plugins/dossier/tests/run.sh rotation-check.test.sh`: 96/96; `shellcheck -S warning -x`: clean.
- [x] All P1/P2/P3 findings in touched files addressed in-PR (or escalated with six-field structure)
- [x] GitHub Actions CI (Linux): 4/4 checks green (Analyze actions, Analyze python, CodeQL, test)

PR #156, opened 2026-08-03
fix(dossier): correct disclosure-gate's rejected-row fixture path and probe sentence

### LINKED ISSUE #151: dossier tests: disclosure-gate.test.sh's rejected-row fixture writes to the wrong path, making that assertion pass vacuously
## Summary

`plugins/dossier/tests/disclosure-gate.test.sh`'s "rejected row is not honoured as approved" test block (around line 288-303) calls `mkpkg "$T"` (which creates `$T/docs/dossier/00-control` and `$T/docs/dossier/06-public`), then writes the fixture's "Rejected and withdrawn claims" table via:

    cat >> "$T/00-control/claim-and-disclosure-register.md"

This is missing the `docs/dossier/` path segment — it should be `$T/docs/dossier/00-control/claim-and-disclosure-register.md`. The heredoc redirect fails (`No such file or directory`), so the register file is never actually written with the rejected-row content, and no register file exists at that fixture path at all.

## Why this matters

The assertion at line 300-303 still reports PASS ("a rejected row is not honoured as an approved claim") — but only because the scan fails for the unrelated reason of a missing register entirely, not because the rejected-row-vs-approved-row logic was actually exercised. This is a vacuously-passing regression test: it provides no real coverage of the behavior it's named for.

Found as a byproduct of PR #150's review (test-runner agent, full-suite run) — unrelated to that PR's diff (this file isn't touched by it), so filed separately rather than folded in.

## Suggested fix

Fix the path in the heredoc redirect to include `docs/dossier/`, matching `mkpkg`'s actual directory layout. After the fix, verify the test still passes — and that it's now passing because the rejected-row logic actually rejected the row, not because the register file was silently absent.

## Acceptance criteria

- [ ] The fixture write path matches `mkpkg`'s actual layout (`$T/docs/dossier/00-control/claim-and-disclosure-register.md`)
- [ ] The "rejected row is not honoured as approved" assertion is verified to fail if the rejected-row filtering logic is temporarily broken (a real RED/GREEN check), not just pass vacuously
- [ ] `bash plugins/dossier/tests/run.sh disclosure-gate.test.sh` passes

danielbentes:
Closes #151.

## Summary

- `disclosure-gate.test.sh`'s "rejected row is not honoured as approved" test wrote its fixture to `$T/00-control/...` instead of `$T/docs/dossier/00-control/...`, so the heredoc redirect failed silently (parent directory didn't exist) and the register was never actually written. The scan fell through the unrelated "no register present" path, and the assertion passed for the wrong reason.
- Fixing the path alone does not satisfy the issue's own acceptance criteria: the probe sentence ("Our service is completely secure.") independently trips the prohibited-vocabulary scanner regardless of registration, forcing exit 1 whether or not the rejected row is honoured as approved. Verified by running the fixed-path version against the real scan script with the rejected-row skip logic both intact and deliberately broken — both cases still exited 1.
- Replaced the probe sentence with mundane wording that trips neither leak nor vocabulary patterns, so exit 0 vs. exit 1 actually distinguishes "wrongly approved" from "correctly rejected."

## Verification

- RED: reverted `dossier-claim-scan.sh`'s rejected-section awk skip logic to always-0 — the redesigned assertion fails as expected.
- GREEN: with production code intact, the assertion passes.
- Full suite: `bash plugins/dossier/tests/run.sh` — 1936 assertions / 0 failures.

## Test plan

- [x] `bash plugins/dossier/tests/run.sh disclosure-gate.test.sh` passes
- [x] Assertion verified to fail when the rejected-row filtering logic is broken (real RED/GREEN, per issue's own AC)
- [x] Full suite run, zero regressions

### DIFF
--- a/plugins/dossier/tests/disclosure-gate.test.sh
+++ b/plugins/dossier/tests/disclosure-gate.test.sh
@@ -289,16 +289,24 @@ assert_not_contains "sk-ant-abcdefgh12345678" "$OUT" "an anthropic key never rea
 # `CL-` rows appear in two tables with different column layouts, and the
 # approval test is a literal substring match. A rejected row whose free-text
 # cell happens to contain the approved marker would otherwise be honoured.
+#
+# The probe sentence is deliberately mundane, NOT "our service is completely
+# secure": that wording independently trips the prohibited-vocabulary scanner
+# (section A, run over the raw document regardless of registration), which
+# forces exit 1 whether or not the rejected row is honoured as approved —
+# confirmed by running both the pass and fail cases below through the real
+# scanner. A sentence with no leak/vocabulary match is required for exit 0 vs.
+# exit 1 to actually distinguish "wrongly approved" from "correctly rejected".
 T="$W/rejected"; mkpkg "$T"
-cat >> "$T/00-control/claim-and-disclosure-register.md" <<'REGEOF'
+cat >> "$T/docs/dossier/00-control/claim-and-disclosure-register.md" <<'REGEOF'
 
 ## Rejected and withdrawn claims
 
 | ID | Wording | Reason declined | Date | What would make it publishable |
 |---|---|---|---|---|
-| CL-9001 | our service is completely secure | the reviewer noted this was not "| approved |" by security | 2026-07-26 | evidence |
+| CL-9001 | our onboarding process takes one business day | the reviewer noted this was not "| approved |" by security | 2026-07-26 | evidence |
 REGEOF
-pub "$T" "Our service is completely secure."
+pub "$T" "Our onboarding process takes one business day."
 scan "$T"
 RC=$?
 if [ "$RC" -eq 0 ]; then
--- a/plugins/dossier/tests/disclosure-gate.test.sh
+++ b/plugins/dossier/tests/disclosure-gate.test.sh
@@ -289,7 +289,6 @@ assert_not_contains "sk-ant-abcdefgh12345678" "$OUT" "an anthropic key never rea
 # `CL-` rows appear in two tables with different column layouts, and the
 # approval test is a literal substring match. A rejected row whose free-text
 # cell happens to contain the approved marker would otherwise be honoured.
-#
 # The probe sentence is deliberately mundane, NOT "our service is completely
 # secure": that wording independently trips the prohibited-vocabulary scanner
 # (section A, run over the raw document regardless of registration), which
--- a/plugins/dossier/tests/disclosure-gate.test.sh
+++ b/plugins/dossier/tests/disclosure-gate.test.sh
@@ -306,13 +306,27 @@ cat >> "$T/docs/dossier/00-control/claim-and-disclosure-register.md" <<'REGEOF'
 |---|---|---|---|---|
 | CL-9001 | our onboarding process takes one business day | the reviewer noted this was not "| approved |" by security | 2026-07-26 | evidence |
 REGEOF
+# A silent failure here is exactly the bug this block was rewritten to fix: a
+# wrong or missing parent directory makes the heredoc redirect fail with the
+# register file never written at all, and the scan below would still exit
+# non-zero (via the unrelated "no register present" path) — passing the
+# assertion for the wrong reason all over again. Checked directly rather than
+# trusted, since that is precisely how the original bug hid.
+assert_file_exists "$T/docs/dossier/00-control/claim-and-disclosure-register.md" \
+  "the rejected-row register was actually written"
 pub "$T" "Our onboarding process takes one business day."
-scan "$T"
+OUT=$(scanout "$T")
 RC=$?
-if [ "$RC" -eq 0 ]; then
-  _dossier_assert_fail "a rejected row was honoured as an approved claim"
-else
-  _dossier_assert_pass "a rejected row is not honoured as an approved claim (exit $RC)"
-fi
+assert_equal "1" "$RC" "a rejected row is not honoured as an approved claim"
+# Pinned individually, not just via the aggregate exit code: exit 1 alone
+# cannot tell "correctly rejected" apart from a handful of other ways this
+# same fixture could produce a non-zero exit for the wrong reason — a
+# register silently absent again (the exact #151 failure shape), a `cd`/mkdir
+# break inside scan()/scanout(), or the probe sentence drifting back toward
+# vocabulary-flagged wording. Each assertion below names exactly one of those
+# and fails on it specifically, rather than letting exit 1 paper over it.
+assert_contains "CLAIM_SCAN_REGISTER_PRESENT=1" "$OUT" "the register was read, not silently absent"
+assert_contains "CLAIM_SCAN_UNREGISTERED_SENTENCES=1" "$OUT" "exactly the rejected-row sentence is unregistered"
+assert_contains "CLAIM_SCAN_PROHIBITED_VOCABULARY=0" "$OUT" "the probe sentence still trips no vocabulary pattern"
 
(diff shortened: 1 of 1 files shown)


danielbentes (commented):
## Review Summary for PR #156 (self-review, Path A paired-reviewer, 10 facets + holdout ×2 lenses)

### Findings: P1: 1, P2: 3, P3: 4

Fixed in-PR: 6. Escalated (deliberately not fixed, reasoning below): 2. Filed as tracked follow-ups (out of scope — different file): 2.

### P1 — Critical
| Finding | Suggested Fix | Confidence | Disposition |
|---|---|---|---|
| F1 `disclosure-gate.test.sh:~313` — the assertion checked only `RC==0` vs. non-zero, so a leak hit (RC=2), infra error (RC=3), or a silently-absent register (RC=1 for the wrong reason) would all read as "pass" — the same vacuous-test shape #151 exists to fix, just relocated. Independently raised by error-skeptic, error-verifier, test-verifier, and security-verifier. | Replaced with `assert_equal "1" "$?"` plus 3 field-level `assert_contains` checks on the scanner's own reported fields (`REGISTER_PRESENT`, `UNREGISTERED_SENTENCES`, `PROHIBITED_VOCABULARY`), each pinning one specific way the exit code alone could be right for the wrong reason. | HIGH | consensus → **fixed** |

### P2 — Should fix
| Finding | Suggested Fix | Confidence | Disposition |
|---|---|---|---|
| F1b (same root cause as F1) — register-write silent failure not guarded | Added `assert_file_exists` on the register path immediately after the heredoc write, before `scan` runs. Verified it fails when the original path bug is reintroduced. | HIGH | consensus → **fixed** |
| F1c (same root cause) — mutation escape: reverting only the probe sentence (not the path) wouldn't be caught | Closed by the new `CLAIM_SCAN_PROHIBITED_VOCABULARY=0` assertion — reverting the sentence now fails this specifically. | HIGH | consensus → **fixed** |

### P3 — Note
| Finding | Suggested Fix | Confidence | Disposition |
|---|---|---|---|
| F2 — commit subject embedded `(#151)`; repo squash-merge config (`COMMIT_OR_PR_TITLE`) uses the commit's own title only for single-commit PRs, and would append `(#156)` regardless, producing a doubled `(#151) (#156)` on `main`. Verified against `gh api repos/.../squash_merge_commit_title` and 5 sampled prior merges (none embed the issue number in the commit subject). | Added a second commit, making this a multi-commit PR — GitHub squash-merge now uses the clean PR title instead of any commit subject. | HIGH | unchallenged → **fixed** |
| F3 — new comment block used a bare `#` paragraph separator, a construct not used elsewhere in this file (17 other multi-line WHY comments flow as one paragraph). | Folded into one paragraph. | HIGH | unchallenged → **fixed** |
| F4 — `pub()` (the document-writing helper, line 13) has no write-success check — same silent-failure shape as the bug this PR fixes, for a different helper. Raised independently by error-skeptic and both holdout-validation lenses (consensus). | **Escalated, not fixed** — see below. | HIGH | consensus → **escalated** |
| F5 — `mkpkg()` (line 12) swallows `mkdir`'s exit status; every downstream write in this fixture depends on it succeeding. | **Escalated, not fixed** — see below. | MEDIUM | unchallenged → **escalated** |

### Escalated for Human Judgment (F4, F5)

**Why not fixed in-PR:** `pub()` and `mkpkg()` are shared helpers used by ~20 other fixtures across this same file, all following the identical no-error-check pattern (pre-existing convention, not introduced by this diff). Adding a check to just the one call site this PR touches would be inconsistent with the rest of the file and would look like an arbitrary one-off rather than a considered pattern. Fixing it properly (auditing and hardening all ~20 call sites, or the helpers themselves) is a materially larger, separate change unrelated to issue #151's actual bug, and the harness's `RUN_TMPDIR`-owned, freshly-created-per-run temp directories make the realistic failure probability very low in practice (no disk-full/permission-denied incidents reported against this suite).

**Risk if left unaddressed:** Low. Would only manifest under disk-full/permission-denied conditions in CI or locally, and would produce a confusing "no register found" or "missing public dir" failure rather than a silent false-pass — i.e., a loud, diagnosable failure, not a silent one, since the file being tested here would then have `assert_file_exists`-class guards fail loudly (as this PR now does for the register write specifically).

**Recommended path:** A dedicated follow-up (file-wide hardening of `mkpkg`/`pub`/write-checking convention across the test suite) if the team wants to close this systematically, rather than folding it into this narrow test-fixture PR.

### Filed as tracked follow-ups (out of scope — different file, `dossier-claim-scan.sh`)
- #158 — rejected/withdrawn heading skip is fail-open on heading-spelling variants (security-verifier)
- #159 — no coverage for a pending-row/free-text-cell approval-marker collision (security-verifier)

### Retracted finding
test-skeptic initially reported a P1 (apparent order-dependent full-suite failure at commit 8994bd2). Independently reproduced by both of us in true isolation (git worktree / git clone, 3 consecutive runs each, both commits): clean every time. Root-caused to concurrent multi-agent test execution against the shared working tree during this review session (confirmed: test-skeptic's first two runs were against the shared tree; all reruns from isolated clones were clean). Retracted — not a PR defect. Process note: the shared working tree is not safe for concurrent test execution by multiple review agents; future multi-agent reviews should use per-agent worktrees/clones.

### Requirements Adherence (issue #151)
| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Fixture write path matches `mkpkg`'s actual layout | ✅ PASS | `disclosure-gate.test.sh:301` writes to `$T/docs/dossier/00-control/...`, matching `mkpkg()` at line 12 |
| 2 | Assertion verified to fail if rejected-row filtering logic is broken (real RED/GREEN) | ✅ PASS | Reverted `dossier-claim-scan.sh`'s awk skip logic 3 separate times during this review (initial fix, post-strengthening, post-holdout) — assertion failed correctly each time; restored and reverified GREEN each time |
| 3 | `bash plugins/dossier/tests/run.sh disclosure-gate.test.sh` passes | ✅ PASS | 48/48, plus 3 consecutive full-suite runs (1931/1931) at current tip |

### Verification
- shellcheck: clean (only pre-existing info-level SC2016 hits on lines this diff doesn't touch)
- CI (Linux): all 4 checks pass (Analyze actions, Analyze python, CodeQL, test) at tip `cb2d6f3`
- Full suite: 3 consecutive clean runs, 1931/1931

danielbentes:
## Self-Review Resolution — PR #156, cycle 1

Fix-forward applied for all fixable findings; two P3s escalated with reasoning (see review comment above), two P3s in an untouched file filed as tracked follow-ups (#158, #159).

**Resolved (fix-forwarded in this PR):**
- F1 (P1) — assertion now checks the exact exit code plus 3 scanner-reported diagnostic fields, not just non-zero
- F1b (P2) — `assert_file_exists` guards the register write itself
- F1c (P2) — `CLAIM_SCAN_PROHIBITED_VOCABULARY=0` closes the sentence-reversion mutation escape
- F2 (P3) — second commit added so squash-merge uses the clean PR title, avoiding a doubled issue-number suffix
- F3 (P3) — comment paragraph style folded to match file convention

**Escalated (not fixed in-PR, reasoning in review comment):**
- F4 (P3) — `pub()` write-check gap, file-wide convention across ~20 fixtures
- F5 (P3) — `mkpkg()` exit-status swallow, same file-wide convention

Commits: 8994bd2 (original fix), c28aaa2 (field-level assertion strengthening), cb2d6f3 (comment style). All RED/GREEN-verified individually and in combination; full suite 1931/1931 across 3 consecutive runs; CI green on Linux.

PR #157, opened 2026-08-03
fix(dossier): close enforce-allowed-actions's quote/backslash/ANSI-C bypass

### LINKED ISSUE #152: dossier: enforce-allowed-actions.sh's BOUND/WRAPPER anchors are bypassed by quoting or backslash-escaping the denied command

danielbentes:
Closes #152.

## Summary

`BOUND`/`WRAPPER`'s boundary anchor is defeated by quoting or backslash-escaping a denied word, independent of any wrapper token. Confirmed live for 8 distinct spellings across 6 underlying bypass mechanisms, each executing the denied command for real:

| Spelling | Technique |
|---|---|
| `"curl" https://evil` | whole word quoted |
| `\curl https://evil` | leading backslash |
| `c\url https://evil` | mid-word backslash split |
| `cu''rl https://evil` | adjacent single-quote fragments (bash joins these into one word) |
| `cu"r"l https://evil` | adjacent double-quote fragments (same mechanism, double quotes) |
| `cur\<newline>l https://evil` | backslash-newline line continuation |
| `$'\143url' https://evil` | bash's own ANSI-C octal quoting |
| `$'\x63url' https://evil` | bash's own ANSI-C hex quoting (same mechanism, hex escape) |

Only the first two were in the original report. The other six (4 spellings across 4 additional mechanisms, 2 of which — quote-fragment joining and ANSI-C — each get 2 tested spellings) were found investigating it: adjacent quote-fragment joining is a known technique (bash treats `r''m` as one token after quote removal — same shape as a recent public exec-allowlist advisory); the newline and ANSI-C cases are bash's own quoting grammar, not a gap specific to this file's regex.

## Fix

Adds `DEQUOTED`: undoes bash's own quote/escape/ANSI-C removal (ANSI-C decoded first since it has its own escape grammar, then backslash-newline joined as a pair before any lone backslash is stripped) before the anchor is tested, so the check runs against the word bash actually executes rather than its disguised spelling. Implemented in bash parameter expansion rather than sed/tr — sed/grep are line-oriented by default and would silently miss the newline-continuation case, and this sidesteps the GNU/BSD sed divergence this suite has hit before.

One bash-3.2-specific wrinkle found mid-fix: macOS's system `/bin/bash` cannot match a `${var//pattern/}` pattern that is itself backslash-followed-by-a-real-newline (confirmed empirically; not reproducible under zsh or newer bash). Worked around with a placeholder byte so the join pattern is never backslash-adjacent.

**Accepted trade-off** (same direction as the file's existing wrapper-scan philosophy): deleting a quote mark can bring a boundary character and a denied word directly together when a quote mark was the entire gap between them (`echo hi;"curl" ...` → `;curl`). Confirmed this is a genuinely new effect of this fix (the pre-fix script permits it), documented in-file, with a test proving it's deliberate.

## Known residual, tracked separately (#160)

Review turned up a categorically different gap: brace expansion (`cu{rl,}`), parameter-expansion defaults (`${x:-curl}`), and command substitution (`$(echo curl)`) all defeat `DEQUOTED` just as completely, each executing the real denied command. Confirmed pre-existing on `main` (not a regression from this PR). Not folded into this fix because it's a different-shaped problem — `DEQUOTED` reverses bash's quote/backslash/ANSI-C *removal*, a deterministic string transform; resolving what a command substitution evaluates to requires either executing attacker-controlled shell content inside the hook (worse than the gap) or full static shell-expansion analysis (not tractable in general). Filed as #160 with live reproductions and documented in-file rather than left implicit.

## Verification

- 9 new assertions in `hooks.test.sh`: one per closed bypass spelling (8) plus the accepted trade-off.
- Each verified RED (fails against the pre-fix script) and GREEN (passes with the fix).
- shellcheck clean.
- Full suite: `bash plugins/dossier/tests/run.sh` — 1936 assertions / 0 failures.

## Test plan

- [x] All 6 quote/backslash/ANSI-C bypass mechanisms (8 tested spellings) now blocked
- [x] All existing wrapper/find-exec/control assertions still pass (no regressions)
- [x] Documented trade-off has a passing, RED/GREEN-verified test
- [x] Full suite run, zero regressions
- [x] CI green on Linux
- [x] Expansion-based residual (brace/parameter-expansion/command-substitution) found in review, confirmed pre-existing, documented in-file, and tracked in #160 rather than left implicit

### DIFF
--- a/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
+++ b/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
@@ -253,6 +253,42 @@ FIND_EXEC='(^|[;&|(]|[[:space:]])[[:space:]]*([A-Za-z0-9_.-]*/)*find([[:space:]]
 # the CI refresh job's own --allowedTools allowlist doesn't include `find`,
 # `xargs`, `parallel`, `awk`, `perl`, or a bare shell at all, so the automated
 # pipeline has no reach here.
+# Bash's \uHHHH/\UHHHHHHHH Unicode escapes (added in bash 4.2) aren't in
+# printf '%b's escape set at all -- confirmed divergence on real bash >=4.2:
+# bash's own $'...' decodes a 4-hex-digit \u escape for codepoint 0x63 to
+# the single byte 'c', but printf '%b' passes that same escape through
+# completely unrecognized, so a denied word spelled this way never reached
+# the deny scan. This was a real, live P1 on CI's Linux bash
+# (which implements \u/\U; the bash on this file's own dev machine does not,
+# so this had to be verified by pushing to CI and reading the failure, not
+# locally). Only the ASCII range (codepoint <= 0x7F) matters for spelling a
+# denied word -- curl/npm/git/etc. are all ASCII -- and UTF-8 encodes that
+# range as the single byte equal to the codepoint, so it's converted to
+# \xHH here, which printf '%b' already handles identically to bash (see the
+# octal note below). A codepoint above 0x7F needs real multi-byte UTF-8
+# encoding to replicate exactly and cannot spell an ASCII denied word
+# regardless, so it's deliberately left unnormalized (passes through as
+# literal text, same as it would if this function didn't exist at all).
+unicode_normalize() { # $1: ANSI-C span body -> stdout: ASCII-range \u/\U normalized to \xHH
+  local body="$1" esc hex cp repl escapes
+  escapes=$(printf '%s' "$body" | grep -oE '\\(u[0-9a-fA-F]{4}|U[0-9a-fA-F]{8})' | sort -u)
+  while IFS= read -r esc; do
+    [ -n "$esc" ] || continue
+    hex="${esc:2}"
+    cp=$((16#$hex))
+    if [ "$cp" -le 127 ]; then
+      repl=$(printf '\\x%02x' "$cp")
+      # $esc must be quoted here: unquoted, its leading backslash is a glob
+      # ESCAPE CHARACTER to `${var//pattern/}` (consumed, not matched
+      # literally), which silently matches one character late and leaves
+      # the original backslash behind duplicated alongside $repl's own --
+      # caught by testing the byte-level output, not by reading the diff.
+      body="${body//"$esc"/$repl}"
+    fi
+  done <<< "$escapes"
+  printf '%s' "$body"
+}
+
 ansi_c_decode() { # $1: command string -> stdout: with every $'...' span decoded
   # Bash's $'...' decodes \nnn (octal), \xHH (hex), and the usual \n/\t/...
   # single-letter escapes. printf '%b' understands the same set, EXCEPT its
@@ -264,6 +300,7 @@ ansi_c_decode() { # $1: command string -> stdout: with every $'...' span decoded
     rest="${s#*\$\'}"
     body="${rest%%\'*}"
     rest="${rest#"$body"\'}"
+    body=$(unicode_normalize "$body")
     norm=$(printf '%s' "$body" | sed -E 's/\\([0-7]{1,3})/\\0\1/g')
     out+="$pre$(printf '%b' "$norm")"
     s="$rest"
--- a/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
+++ b/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
@@ -271,6 +271,30 @@ ansi_c_decode() { # $1: command string -> stdout: with every $'...' span decoded
   printf '%s' "$out$s"
 }
 
+# ansi_c_decode forks a sed+printf subprocess pair for every $'...' occurrence
+# it finds. The byte-length guard above bounds the quadratic glob-matching
+# cost, but not this one: an attacker can still pack many tiny $'x' spans
+# into that same byte budget, and fork/exec overhead is paid per span
+# regardless of how small each one is -- confirmed live, ~2000 minimal spans
+# in an 8KB command took over 10 seconds from subprocess overhead alone, with
+# no `$'...'`-external content large enough to trip the length guard. A
+# legitimate interactive command needs at most a handful of ANSI-C-quoted
+# segments, so this is capped separately from the overall length.
+#
+# Counted with `grep -oF`, not bash's own `${var//pattern/replacement}`:
+# bash's global substitution has the exact same super-linear-for-many-matches
+# behavior this guard exists to bound -- confirmed live, counting via pure
+# bash parameter expansion on the same 2000-span input took ~7 seconds by
+# itself, which would have defeated the point of counting fast before
+# deciding whether to run the expensive decode. `-F` matches the 2-character
+# literal `$'` identically on GNU and BSD, so this doesn't reopen the
+# GNU/BSD divergence risk the newline-join logic above avoids sed/grep for.
+ANSIC_SPAN_COUNT=$(printf '%s' "$COMMAND" | grep -oF "\$'" | wc -l | tr -d '[:space:]')
+if [ "$ANSIC_SPAN_COUNT" -gt 64 ]; then
+  echo "BLOCKED: dossier cannot safely enforce its action ceiling on a command with this many \$'...' segments (${ANSIC_SPAN_COUNT}, limit 64) -- each one forks a subprocess to decode." >&2
+  exit 2
+fi
+
 COMMAND_ANSIC=$(ansi_c_decode "$COMMAND")
 
 # The placeholder byte chosen below (0x01) is assumed to never legitimately
--- a/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
+++ b/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
@@ -271,6 +271,29 @@ ansi_c_decode() { # $1: command string -> stdout: with every $'...' span decoded
   printf '%s' "$out$s"
 }
 
+# Three known grammar divergences between ansi_c_decode and bash's own
+# $'...' parsing, each reviewed and confirmed non-exploitable rather than
+# left unmentioned:
+#   - `\'` inside a span (`$'cur\'l'`): bash decodes this to the 5-char
+#     string `cur'l` (the escaped quote is literal, decoding continues past
+#     it); this function's span-boundary search stops at the FIRST literal
+#     `'` regardless, so it mis-locates where the span ends. Not exploitable:
+#     any use of `\'` inside a span leaves a literal quote character in
+#     bash's own real decoded word too (`cur'l`, not `curl`), so it can never
+#     spell a clean, standalone denied word on the side that actually runs.
+#   - `\143` immediately after a literal backslash (`\\143`): the sed
+#     octal-normalization step below treats the second backslash as if it
+#     started a fresh octal escape, diverging from bash's own 4-byte decode
+#     of the same input. Same non-exploitability argument: the divergence
+#     only ever surfaces a stray backslash in bash's real decoded word too.
+#   - `\c` (control-character escape): bash's `\cX` consumes exactly the next
+#     character to form one control byte and keeps decoding after it; POSIX
+#     printf's `\c` means "stop all further output" instead, a completely
+#     different operation. Not exploitable: `\cX` always glues an
+#     unprintable control byte to whatever it's adjacent to, which corrupts
+#     the resulting word identically for bash's real execution (the command
+#     bash actually looks up on PATH also stops being a clean "curl"/"npm"/
+#     etc.), not just for this function's decode.
 # ansi_c_decode forks a sed+printf subprocess pair for every $'...' occurrence
 # it finds. The byte-length guard above bounds the quadratic glob-matching
 # cost, but not this one: an attacker can still pack many tiny $'x' spans
@@ -319,14 +342,17 @@ case "$COMMAND_ANSIC" in
     ;;
 esac
 
-# bash 3.2 (macOS's system /bin/bash) cannot match a pattern-substitution
-# pattern that is itself a literal backslash followed by a real newline: the
-# 2-character pattern's own length checks out, but the replacement silently
-# never fires -- confirmed empirically, and NOT reproducible under zsh or
-# newer bash, which both do fire. A backslash is a glob escape character
-# inside `${var//pattern/}`, and bash 3.2 mishandles escaping a newline
-# specifically. Every backslash is swapped for a placeholder byte first (one
-# that never legitimately appears in a Bash-tool command string) so the
+# bash 3.2 (macOS's system /bin/bash) mishandles a backslash-followed-by-
+# real-newline pattern specifically when that 2-character pattern is
+# pre-assembled into a variable before being used in `${var//pattern/}` --
+# the substitution then silently never fires, even though the identical
+# pattern written inline (`${var//\\$'\n'/}`) matches correctly on the same
+# bash. Confirmed empirically on both forms; NOT reproducible under zsh or
+# newer bash, where both forms fire. The placeholder-byte approach below
+# sidesteps the question of which form is safe on which bash version
+# entirely, rather than depending on getting that distinction right: every
+# backslash is swapped for a placeholder byte first (one that never
+# legitimately appears in a Bash-tool command string, checked above) so the
 # join pattern below is placeholder+newline, never backslash+newline; the
 # remaining lone placeholders are then deleted, which is step 3 (the escape
 # character removed, the character it protected kept) applied via the same
--- a/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
+++ b/plugins/dossier/hooks/scripts/enforce-allowed-actions.sh
@@ -67,6 +67,24 @@ if [ "$JQ_RC" -ne 0 ]; then
 fi
 [ -z "$COMMAND" ] && exit 0
 
+# Fail closed rather than let bash's own glob-based parameter expansion pay a
+# super-linear cost: ansi_c_decode's span-boundary search (`${rest%%\'*}`,
+# re-scanning a shrinking-but-still-large remainder every loop iteration) and
+# the backslash-to-placeholder pass further below (a single global
+# substitution over the whole command) are both quadratic in bash for large
+# operands -- confirmed empirically: a 100KB command containing ANSI-C
+# content took ~20s to normalize here, ~200KB took over a minute, and the
+# same blowup reproduces from backslash volume alone with no `$'...'`
+# involved at all. A command this large is never a legitimate interactive
+# Bash-tool call this hook needs to support, so refusing it outright costs
+# nothing real while closing an easy, no-malice-required hang -- and this
+# hook blocks the whole Bash tool call while it runs, not just itself.
+COMMAND_LEN=${#COMMAND}
+if [ "$COMMAND_LEN" -gt 8192 ]; then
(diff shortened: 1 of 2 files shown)


danielbentes (commented):
## Review summary (Path A, paired-reviewer + self-review fix-forward)

**Facet coverage note (transparency):** convention, test, and error-handling facets received both skeptic and verifier passes and are fully consolidated below. Security and code-quality facets received one reporting lens each (security-skeptic; code-verifier) — the paired counterpart did not report back in this session. Rather than leave those facets single-lens, I ran my own adversarial pass over the diff's only new executable code (`ansi_c_decode`) and independently reproduced/verified the security-skeptic's finding before acting on it. Findings from single-lens facets are marked `unchallenged` below per the paired-reviewer protocol's fallback rule, not `consensus`/`validated`.

### Findings

| ID | Priority | Category | Location | Status | Disposition |
|---|---|---|---|---|---|
| F1 | P1 | convention | commit history (squash-title risk) | **Fixed** | validated |
| F2 | P2 | convention | `enforce-allowed-actions.sh:153-164` (count accuracy) | **Fixed** | validated |
| F3 | P3 | convention | `enforce-allowed-actions.sh:171-190` (list formatting) | **Fixed** | validated |
| F4 | — | security | brace/param-expansion/command-substitution defeat `DEQUOTED` | **Tracked** (#160), documented in-file | unchallenged (single-lens; independently reproduced by me on both branches) |
| F5 | — | test | issue #152's own 3rd reproduction (`\find . -exec curl {} \;`) had no dedicated assertion | **Fixed** — assertion added, RED/GREEN verified | self-found during requirements review |
| F6 | P3 | error-handling | `ansi_c_decode`'s internal `sed`/command-substitution has no explicit failure check | Noted, not fixed | not practically triggerable — `sed` does not fail on a valid substitution over arbitrary text; theoretical only |

No P1/P2 remains open. F4 is a pre-existing, confirmed-non-regression gap (checked against `main` before `#152`) that is a categorically different problem (shell *expansion*, not quote/escape *removal*) — tracked with live reproductions rather than silently deferred, per this file's own precedent for gaps with "no concrete trigger yet."

### Adversarial pass on new code (my own, filling the missing code-skeptic lens)

`ansi_c_decode` is the only new executable code in this PR. Checked, live, with hang-detection and a real curl shim on `PATH`:
- Unterminated `$'...'` (no closing quote): no hang (resolves in ~1s), decoder leaves it unchanged. Confirmed real bash itself rejects this as a syntax error before ever executing anything, so it isn't a viable input regardless of the decoder's behavior.
- Hostile content inside `$'...'` (`\'`, `\c`, `%s%s%s`): no crash. `%`-sequences are passed as `printf '%b'`'s *argument*, never its format string, so no format-string injection. Bash's `\cX` control-escape and printf's `\c` "stop output" escape diverge semantically, but every combination I could construct that exploited the divergence also corrupted the resulting word for bash's *own* real execution — no path found where the hook is fooled while bash still runs the clean, exact denied word.

### Requirements adherence (issue #152)

| Requirement | Status |
|---|---|
| Close `"curl"` (whole-word-quoted) bypass | ✅ closed, tested |
| Close `\curl` (leading-backslash) bypass | ✅ closed, tested |
| Close `\find . -exec curl {} \;` (backslash + find/-exec combined) | ✅ closed — **test gap found and fixed in this review** (F5) |
| "Normalize quotes/backticks/backslashes/`$(`... unconditionally" (issue's suggested fix, verbatim) | Quotes/backslash/backtick/ANSI-C: ✅ implemented unconditionally, before any WRAPPER test. `$(`: **deliberate, verified deviation** — empirically confirmed that naively stripping `$(`/`)` characters (the issue's literal suggestion) would not have closed anything (`echo curl https://evil` still doesn't match the boundary anchor), so it was not implemented that way. Documented in #160 rather than shipped as a no-op fix. |
| No per-token enumeration | ✅ `DEQUOTED` is a blanket transform, not a per-word list |

### Holdout validation

Evaluated against behavioral and error-handling scenario sets (scenario IDs only, per protocol — specifics withheld).

- Behavioral: scenario-1 through scenario-5 — PASS. New assertions verify actual exit-code/blocking behavior (not existence-only), cover 9 distinct spellings plus the trade-off case (not happy-path-only), all have real expected-vs-actual assertions, all confirmed current against the live file (re-run, 108/108), and all test through the hook's real external interface (stdin JSON → exit code), not internal state.
- Error-handling: scenario-1 through scenario-5 — PASS, with one informational note (F6 above) on an unchecked-but-not-practically-triggerable internal command substitution.
- No conflicts between self-review claims and file state found beyond what's captured in F1-F6 above.

### Verification

- Full suite: `bash plugins/dossier/tests/run.sh` — 1937 assertions / 0 failures (was 1936 before F5's new assertion).
- `hooks.test.sh` alone: 108/108 (was 107; +1 from F5).
- shellcheck: clean.
- CI (this tip): CodeQL, test, Analyze (actions), Analyze (python) — all green.
- PR #156 (bundled review): re-verified full suite 2× in an isolated `git worktree` at its current tip (`cb2d6f3`) — 1931/1931 both runs, no flake recurrence.

danielbentes:
## Self-review resolution (cycle 1)

- **F1** (P1, squash-title double-suffix risk): resolved — branch now carries 2 commits, so GitHub's `squash_merge_commit_title=COMMIT_OR_PR_TITLE` falls back to the PR title rather than the first commit's `(#152)`-suffixed subject.
- **F2** (P2, "six"/8-spelling count accuracy): resolved — in-file comment and PR body corrected.
- **F3** (P3, numbered-list formatting): resolved — reformatted to match `detect-local-merge.sh`'s precedent.
- **F5** (test gap, issue #152's own 3rd repro case untested): resolved — dedicated assertion added, RED/GREEN verified.

**Escalated (not fixed in this PR):**
- **F4** (brace/parameter-expansion/command-substitution bypass class): confirmed live on both `main` (pre-#152) and this branch — pre-existing, not a regression. Categorically different problem from what `DEQUOTED` solves (shell *expansion* vs quote/escape *removal*); a real fix requires either executing attacker-controlled content inside the hook (worse than the gap) or full static shell-expansion analysis (not tractable in general). Filed as #160 with live reproductions, a documented dead-end (the issue's own literal "strip `$(`" suggestion was verified not to work), and candidate approaches for whoever picks it up. Documented in-file alongside the existing accepted-residual precedent (`parallel`/`awk`/`perl`).
- **F6** (P3, unchecked internal `sed`/command-substitution in `ansi_c_decode`): not fixed — verified not practically triggerable (`sed` does not fail on a valid substitution over arbitrary text). Noted for awareness, not tracked as a separate issue given no realistic trigger.

**Disputed:** none.

danielbentes (commented):
## Review cycle 2 — retraction and correction

Cycle 1's review comment stated "No P1/P2 remains open" and its facet-coverage note claimed security and code-quality had only single-lens (unchallenged) coverage. **That was wrong on both counts.** All ten paired-reviewer lenses reported (they simply took longer than I waited for), and they found three real, live, independently-reproduced P1s in this PR's own new code. Retracting cycle 1's "no P1" claim and its coverage-gap framing. This cycle supersedes it.

### New findings (all fixed, all independently verified by me before and after the fix)

| ID | Priority | Category | Location | Status | Disposition |
|---|---|---|---|---|---|
| F7 | P1 | security | `enforce-allowed-actions.sh` — `DEQUOTE_PLACEHOLDER` (0x01) collision | **Fixed** | consensus (error-skeptic + code-skeptic both found the collision independently; error-skeptic additionally proved the exploit path) |
| F8 | P1 | performance/security | `enforce-allowed-actions.sh` — quadratic-time hang in `ansi_c_decode`'s glob-matching | **Fixed** | consensus (error-verifier + security-verifier both measured it independently, matching orders of magnitude) |
| F9 | P1 | performance/security | `enforce-allowed-actions.sh` — per-span fork/exec overhead, unbounded by the byte-length fix for F8 | **Fixed** | self-found verifying F8's fix (a naive bash-native counting approach for the guard turned out to have the *same* quadratic pathology as F8 — caught before it shipped) |

**F7 detail:** `ansi_c_decode` can manufacture the literal 0x01 byte from a plain `$'\001'` escape in the original command — the exact byte `DEQUOTE_PLACEHOLDER` uses internally. That byte is then indistinguishable from a backslash the script marks itself; if it sits before a real newline, the join step deletes both, splicing two statements together with no boundary character for `BOUND`/`BOUND_ACTIVE` to anchor on. Confirmed live: `echo done$'\001'` on one line followed by a real `curl ...` on the next was permitted outright before the fix. Fixed by failing closed whenever the ANSI-C-decoded command already contains that byte (covers both this manufactured case and a literal 0x01 typed directly into the command).

**F8 detail:** `ansi_c_decode`'s span-boundary search (`${rest%%\'*}`) and the backslash-to-placeholder pass both re-scan a shrinking-but-still-large remainder, which is quadratic in bash for large operands. Confirmed live: 100KB of ANSI-C content took ~20s to normalize, ~200KB took over a minute — the whole Bash tool call blocks for that duration. Fixed with an early byte-length guard (8192 bytes) before any of the expensive normalization runs.

**F9 detail:** found verifying F8's fix. The byte-length guard doesn't bound fork/exec overhead: `ansi_c_decode` forks a `sed`+`printf` pair per `$'...'` occurrence, so many *tiny* spans stay under the byte cap while still paying that cost per span — confirmed live, ~2000 minimal spans in an 8KB command took over 10s. My first attempt at a fix (counting spans via bash's own `${var//pattern/replacement}`) turned out to have the *identical* quadratic-for-many-matches behavior — confirmed the counting step alone took ~7s on the same input before I caught it and switched to `grep -oF` (a single fork, ~0.01s, correctness-equivalent for a fixed 2-character literal on both GNU and BSD).

### Other findings from this cycle — triaged, not all fixed

| Finding | Disposition |
|---|---|
| test-verifier: `\u`/`\U` Unicode ANSI-C escapes unnormalized | **Addressed with a CI-verifying test**, not a code fix — bash 3.2 (local dev bash) doesn't implement `\u`/`\U` in either `$'...'` or `printf '%b'` at all, so this can't be resolved locally. Added a bash>=4.2-gated assertion (`hooks.test.sh`) that will actually exercise the divergence question on Linux CI's bash. If it fails there, that's a live P1 to fix before merge, not after — watching `gh pr checks` on this push. |
| test-verifier: WRAPPER-token-disguised (not just payload) untested | **Fixed** — added a `\xargs -I{} curl {}` regression test; verified still-correct behavior, now proven by the suite. |
| test-skeptic: second (quoted-wrapper) over-block trade-off undocumented/untested | **Fixed** — added a regression test (`;"env" grep -r "npm test" docs/`) and a code comment, matching how the first trade-off is already documented. |
| code-skeptic: bash-3.2 backslash-newline claim overstated | **Fixed** — verified the narrower actual constraint myself (inline pattern form fires correctly; only a pre-assembled-pattern-in-a-variable form fails) and rewrote the comment to the narrower, now-verified claim. |
| code-skeptic + security-verifier: `\'`-inside-span mis-parse, `\\143` sed mis-tokenization, `\c` truncation | **Documented, not fixed** — three independent reviewers (plus my own check) converged on non-exploitable: each divergence corrupts the word bash *itself* actually executes too, not just this function's decode, so no construction spells a clean denied word on the executing side while hiding from the hook. Added an explicit code comment naming all three so a future refactor doesn't have to re-derive the argument. |
| error-skeptic: unchecked `sed`/`printf` exit codes inside `ansi_c_decode` | **Not fixed, not tracked** — the reporter's own testing found no case where keyword letters were dropped by a malformed escape, and F8/F9's guards now bound the input these calls ever see. Adding fail-closed paths for a failure mode nobody could trigger is scope creep the fix doesn't need. |
| test-verifier: fragment-join proven only for N=2/single quote-type; no whole-word single-quote case | **Not added** — the deletion-only mechanism argument (three independent reviewers made it) covers N-way and mixed-type joins identically; low marginal value for the assertion count already in place. |

### Verification

- Full suite: `bash plugins/dossier/tests/run.sh` — 1946 assertions / 0 failures (was 1937 at the end of cycle 1).
- `hooks.test.sh` alone: 117/117 (was 112 at end of cycle 1; +5 net: 4 new RED/GREEN-verified assertions for F7-F9, +3 for the triaged findings, -2 that don't apply on this local bash version and are version-gated).
- Every new assertion for F7/F8/F9 individually verified RED (fails against the pre-fix commit, for the stated reason) and GREEN (passes with the fix) before being committed.
- shellcheck: clean.
- Pushed at `e0d0944`; CI run in progress — will confirm green (including the bash>=4.2-gated `\u` assertion, which only executes meaningfully on Linux) before considering this ready to merge.

danielbentes:
## Self-review resolution (cycle 2)

**Retracting cycle 1's resolution marker's implicit claim of completeness.** Three P1s (F7, F8, F9) landed after cycle 1's marker was posted, from lenses that reported after I'd stopped waiting for them. All three are fixed and verified below.

- **F7** (P1, DEQUOTE_PLACEHOLDER 0x01 collision): resolved — fail-closed guard added, RED/GREEN verified, live-reproduced before and after.
- **F8** (P1, quadratic-time hang from glob-matching on large operands): resolved — byte-length guard (8192) added before any expensive normalization runs.
- **F9** (P1, per-span fork/exec overhead unbounded by F8's fix): resolved — span-count guard (64) added via `grep -oF`, after discovering my first attempt (bash-native counting) had the same quadratic pathology as F8.
- Test-verifier's WRAPPER-token-disguised gap and test-skeptic's quoted-wrapper trade-off: resolved — regression tests added for both.
- Code-skeptic's bash-3.2 comment overstatement: resolved — comment tightened to the verified, narrower constraint.

**Escalated (unchanged from cycle 1):**
- F4 (brace/parameter-expansion/command-substitution bypass class): tracked in #160, pre-existing and not a regression.

**Escalated (new, contingent on CI):**
- test-verifier's `\u`/`\U` Unicode-escape divergence hypothesis: cannot be resolved on local bash 3.2 (doesn't implement `\u`/`\U` at all). Added a bash>=4.2-gated test that will run for real on Linux CI. If CI is green, this is closed with evidence. If it fails, it's a P1 to fix before merge, not after — not merging until this is confirmed either way.

**Documented, not fixed (three independent reviewers converged on non-exploitable):**
- `\'`-inside-span mis-parse, `\\143` sed mis-tokenization, `\c` truncation — each corrupts the word bash itself actually executes too, so no bypass. Comment added naming all three.

**Not fixed, not tracked (reporter's own evidence found no live trigger):**
- Unchecked `sed`/`printf` exit codes inside `ansi_c_decode` (error-skeptic P2).

**Disputed:** none.

danielbentes (commented):
## Review cycle 3 — the `\u`/`\U` escaped question resolved (real bug, fixed)

Cycle 2's escalated item (`\u`/`\U` Unicode ANSI-C escapes) came back from Linux CI as a **confirmed failure**. Investigated rather than filed:

- **The underlying gap is real.** `ansi_c_decode`'s `printf '%b'` doesn't understand bash's `\uHHHH`/`\UHHHHHHHH` escapes (added in bash 4.2) at all. Verified directly: a command where the *entire invoked word itself* is spelled via `\u`-escape (e.g. `$'curl' https://example.invalid`) was permitted before this fix and correctly denied after, on both the pre-fix commit and the fix — confirmed by testing both directly, not by trusting one CI result.
- **My first CI-failing test had its own bug.** It prefixed the escaped word with `echo `, which is *always* permitted regardless of decoding (`echo` never invokes its arguments) — so that specific failure wasn't proof of the real gap, just a malformed assertion that happened to also fail. Caught this by re-deriving pre-fix/post-fix behavior directly against the corrected command shape rather than accepting the first CI signal at face value.
- **Fixed with pure string parsing**, not native bash `\u` support: the ASCII range (codepoint ≤ 0x7F, which is all that matters for spelling `curl`/`npm`/etc.) is converted to `\xHH` before printf sees it, the same way octal escapes already are. Above 0x7F is left unnormalized (can't spell an ASCII denied word, and real multi-byte UTF-8 encoding would be needed to replicate exactly). Because this is manual parsing rather than reliance on the runtime bash's own escape grammar, it works identically on bash 3.2 (this dev machine) and bash ≥5 (CI) — confirmed by running the fix locally despite this machine never having native `\u` support to begin with. The regression tests run unconditionally (no version gate needed after all).

### Verification

- Full suite: `bash plugins/dossier/tests/run.sh` — 1948 assertions / 0 failures.
- Both new assertions verified RED (pre-fix commit, `expected '2', got '0'`) and GREEN.
- shellcheck: clean.
- Pushed at `7f9f6a9` — waiting on `gh pr checks` (this is the run that actually confirms the fix on the platform where the original failure was observed).

danielbentes:
## Self-review resolution (cycle 3)

- **U1** (`\u`/`\U` Unicode ANSI-C escape, escalated in cycle 2 pending CI): resolved. Confirmed real (not a pre-existing/tracked residual like #160 — a genuine gap in this PR's own new `ansi_c_decode`), fixed via ASCII-range normalization to `\xHH`, verified RED/GREEN against both script versions directly. My first test attempt had its own construction bug (an `echo ` prefix that made the assertion meaningless regardless of the real gap) — corrected and verified the fix using the right command shape before concluding anything.

**All escalations from cycles 1-3 now resolved except:**
- F4 (brace/parameter-expansion/command-substitution): tracked in #160, pre-existing, not a regression, not fixable as a quick patch (would require executing attacker-controlled content).

**Disputed:** none.

